Issue #75, opened 2019-12-05
Write serializer for KokkosCrsMatrix


lifflander:

lifflander:
Merged, closing

uhetmaniuk:
@lifflander @PhilMiller 
Is there a test for serializing a Kokkos::CrsMatrix variable?

In EMPIRE I serialized / deserialized one Kokkos::CrsMatrix variable, everything looks correct except the number of columns.
In the output of deserialization, the number of columns was incorrectly set to 0.

Please advise whether I should reproduce this issue outside of EMPIRE-PIC

PR #76, opened 2019-12-06
#75: Add serialization support for Kokkos StaticCrsGraph and CrsMatrix

PhilMiller:


PhilMiller:
I have the corresponding cmake hack to let the tests compile with Kokkos OpenMP on branch 75-kokkos-crsmatrix-with-cmake-hack

We need to patch that up to use the right variables from Kokkos rather than hard-coding it. I don't think those tests are built in EMPIRE, though, so we can integrate this and pull it downstream without that being fixed.

PhilMiller:
As to relevant variables:

From `KokkosConfig.cmake`, there's 

```cmake
SET(Kokkos_CXX_FLAGS "     --std=c++11 -fopenmp     -O3 -DNDEBUG")
SET(Kokkos_C_FLAGS "    -fopenmp -O3 -DNDEBUG")
```

From `kokkos_generated_settings.cmake` there's

```cmake
set(KOKKOS_CXXFLAGS --std=c++11 -fopenmp CACHE STRING "" FORCE)
set(KOKKOS_LDFLAGS -fopenmp CACHE STRING "" FORCE)
set(KOKKOS_CXXLDFLAGS CACHE STRING "" FORCE)
set(KOKKOS_LINK_FLAGS -fopenmp CACHE STRING "extra flags to the link step (e.g. OpenMP)" FORCE)
```

In any of those cases, we'd need to extract just the relevant OpenMP flag, and maybe leave the others behind - certainly the `--std=*` bit, since we need to compile with C++14.

As an alternative, we could look for OpenMP here

```cmake
set(KOKKOS_GMAKE_DEVICES "OpenMP,Serial" CACHE STRING "Kokkos devices list" FORCE)
```

And then ask CMake to do the right thing for the compiler in use. @mperrinel do you know how to make that happen?

PhilMiller:
I think this is what we want to end up using:
https://cmake.org/cmake/help/v3.11/module/FindOpenMP.html

examples/CMakeLists.txt
```diff
@@ -39,7 +39,19 @@ if (${CHECKPOINT_BUILD_EXAMPLES})
 
     target_link_libraries(
       ${EXAMPLE}
+      PUBLIC
```
PhilMiller, line 42: I had to add this so CMake wouldn't complain about mixing plain and keyword syntax. It's in a linked example, so I think it's no big deal how we do it.

examples/CMakeLists.txt
```diff
@@ -39,7 +39,19 @@ if (${CHECKPOINT_BUILD_EXAMPLES})
...
       ${SERDES_LIBRARY_NS}
     )
+
+  if (${kokkos_DIR_FOUND})
+    target_include_directories(${EXAMPLE} SYSTEM PUBLIC ${Kokkos_INCLUDE_DIRS})
+    target_link_libraries(${EXAMPLE} PUBLIC ${Kokkos_LIBRARIES})
+
+      find_package(OpenMP)
```
lifflander, line 50: This is definitely not the ideal way to do this, but it works for now.

lifflander (approved):

PR #85, opened 2020-04-20
#84: add traits class to determine whether type is byte copyable or not

### LINKED ISSUE #84: isByteCopyable trait class
Right now the only way to specify a type is byte copyable is by defining a data member type isByteCopyable that is identical to std::true_type

Non-intrusive serialization can be done by making a traits class isByteCopyable and then making it specializable. This would also reduce the compile time cost of using the detector idiom.

nmm0:
Fixes #84


lifflander (approved):
Looks good to me. The detection probably doesn't need covertable, but it doesn't hurt.

PR #87, opened 2020-04-28
#86 Write example checkpointing bytes to disk

### LINKED ISSUE #86: Add example checkpointing bytes to file
Write a small example per request here:

https://cee-gitlab.sandia.gov/EM-Plasma/EMPIRE-PIC/issues/719

lifflander:
Fixes #86

### DIFF
--- a/examples/checkpoint_file.cc
+++ b/examples/checkpoint_file.cc
@@ -1,0 +1,171 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                             checkpoint_file.cc
+//                           DARMA Toolkit v. 1.0.0
+//                 DARMA/checkpoint => Serialization Library
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
+// Government retains certain rights in this software.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+//   this list of conditions and the following disclaimer.
+//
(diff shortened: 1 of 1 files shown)


PhilMiller:
Serializing everything into a big in-memory buffer, and then writing it out, or doing the reverse on read, isn't going to fly. That means memory footprint during checkpoint or resume will be roughly double the live data. If a job is already filling more than half of available memory just to run, then this will fail.

lifflander:
> Serializing everything into a big in-memory buffer, and then writing it out, or doing the reverse on read, isn't going to fly. That means memory footprint during checkpoint or resume will be roughly double the live data. If a job is already filling more than half of available memory just to run, then this will fail.

Yes, agreed. This was meant to be just a very simple example of how one might write to disk. We should design the API for checkpoint to do this automatically.

PhilMiller:
I realized earlier today that the right implementation might be really easy to bolt into the memory buffer model. 

* Do the sizing as usual
* Open a file
* (if packing) `fallocate` that file to the right size
* `mmap` the file
* pass the address to the normal memory buffer serializer
* (optionally call `madvise(MADV_WONTNEED)` every few megabytes)

This lets the kernel and filesystem take care of things, and somewhat naturally paces operation to the filesystem's throughput without really blocking anywhere.

PhilMiller:
Just to be clear, though, the EMPIRE folks need a demonstration *without obvious holes to poke in it*

lifflander:
@PhilMiller Ok, let's spin off some issues to get checkpoint ready for real, to-disk checkpoints. 

Also, we might want some index to write multiple objects to disk in the same file that are serialized independently. I don't know if there's a single root object that can be the singular source.

lifflander:
Closing as this is replaced by a real API!

Take a look at #92

PR #90, opened 2020-05-05
89 Major cleanup

### LINKED ISSUE #89: Cleanup namespace and file names

lifflander:
Fixes #89 

- Fix old file names with "serdes" before this library was called checkpoint
- Fix relative path cmake hack so semi-absolute paths are required for header files
- Fix installation paths for checkpoint headers
- Make everything in namespace "checkpoint"
- Remove all the now unused parserdes
  - Partial serialization was an optimization experiment where byte copyable members weren't serialized just non-byte copyable and then the struct was serialized/sent with non-byte-copyable members reconstituted 
- Use proper API for the examples instead of calling the internal API
- Put the private API interface into its own namespace to reduce confusion `checkpoint::dispatch`. The functions `checkpoint::dispatch::serializeType` and `checkpoint::dispatch::deserializeType` are not supposed to be used as an external API. `checkpoint::serialize` and `checkpoint::deserialize` are the appropriate functions for external use.
- Write proper documentation for the checkpoint API


examples/checkpoint_example_1.cc
```diff
@@ -91,14 +91,14 @@ int main(int, char**) {
   my_test_inst.a = 10;
   my_test_inst.print();
 
-  auto serialized = checkpoint::serializeType<MyTest>(my_test_inst);
+  auto ret = checkpoint::serialize<MyTest>(my_test_inst);
```
PhilMiller: Not something for a cleanup PR, but why does this function require the template argument to be specified, rather than inferred?

src/checkpoint/checkpoint_api.h
```diff
@@ -51,29 +51,93 @@
...
+/**
+ * \brief Serialize \c T into a byte buffer
+ *
+ * Serializes an object recursively by invoking the \c serialize
+ * functions/methods recursively.
+ *
+ * \param[in] target the \c T to serialie
+ * \param[in] fn (optional) callback to supply buffer for to allow user
+ * allocation of the produced byte buffer. The callback will pass the number of
+ * bytes required and return a char* will at least that many bytes.
```
PhilMiller: Typo 'will' -> 'with'

The text should be clarified to read

" The callback will be passed the number of bytes required and return a char* to a buffer of at least that many bytes."

src/checkpoint/checkpoint_api.h
```diff
@@ -51,29 +51,93 @@
...
+ * \param[in] user_buf (optional) buffer containing bytes allocated with
+ * sufficient size for \c T. If this buffer is passed, the caller is responsible
+ * for deallocating the buffer. If it is not passed, the system will allocate a
+ * buffer that must be de-allocated with \c delete
+ *
+ * \return a pointer to the newly reified \c T based on bytes in \c buf
+ */
 template <typename T>
-T* deserialize(char* buf, std::size_t size, T* user_buf = nullptr);
+T* deserialize(char* buf, std::size_t len, T* user_buf = nullptr);
```
PhilMiller: Would it make sense to have two distinct overloads here
```
T* deserialize(char* data_buf, std::size_t len, char* object_buf);
std::unique_ptr<T> deserialize(char* buf, std::size_t len);
```
to distinguish the ownership cases?

I think the second case is fully sound.

In the first case, I'm conflicted as to what type `object_buf` should actually have, and whether the function should return `T*` or `void`. At the point of the call, `object_buf` is not a "pointer to `T`", which would argue for accepting `char*` or `void*`. Returning `T*` would be a convenience to avoid ceremony in a call like

`T* t = deserialize(buf, len, malloc(sizeof(T)));`

But would feel weird in the case of 

`deserialize(buf, len, &array_of_t[i]);`

PhilMiller: Why does this take a `len` argument at all? I know the serializer could ultimately check that it doesn't read off the end of the buffer, and error out or throw an exception, but does it?

src/checkpoint/checkpoint_api.h
```diff
@@ -51,29 +51,93 @@
...
+/**
+ * \brief Convenience function for de-serializing and reify \c T directly from \c
+ * in the return value from \c serialize
+ *
+ * \param[in] in the buffer and size combo returned from \c serialize
+ *
+ * \return a pointer to \c T that must be deallocated
+ */
 template <typename T>
 T* deserialize(SerializedReturnType&& in);
```
PhilMiller, line 148: Why is this taking its argument by r-value reference? It doesn't "consume" the buffer in any way that I could see - there might be reasons to want to deserialize multiple copies of the same object.

src/checkpoint/checkpoint_api.h
```diff
@@ -51,29 +51,93 @@
...
-void deserializeInPlace(char* buf, std::size_t size, T* t);
+void deserializeInPlace(char* buf, std::size_t len, T* t);
 
+/**
+ * \brief Convenience function for de-serializing and reify \c T directly from \c
+ * in the return value from \c serialize
+ *
+ * \param[in] in the buffer and size combo returned from \c serialize
+ *
+ * \return a pointer to \c T that must be deallocated
```
PhilMiller, line 145: Again this should return `std::unique_ptr<T>`

PhilMiller (approved):

examples/checkpoint_example_1.cc
lifflander: It doesn't actually require the template parameter. Not reason to have it.. will remove.

lifflander: Removed from all examples

src/checkpoint/checkpoint_api.h
lifflander: Fixed

src/checkpoint/checkpoint_api.h
lifflander: You are right, it doesn't actually need that.

src/checkpoint/checkpoint_api.h
lifflander: Because it doesn't move the return type out, it's unusable after calling this. I agree that doesn't have to be the semantic if `SerializedReturnType` wasn't a unique_ptr.

src/checkpoint/checkpoint_api.h
lifflander: Ok, added a unique_ptr interface

src/checkpoint/checkpoint_api.h
```diff
@@ -94,16 +95,32 @@ SerializedReturnType serialize(T& target, BufferCallbackType fn = nullptr);
  * deallocated with \c delete
  *
  * \param[in] buf the buffer containing the bytes to reify \c T
- * \param[in] len the number of bytes in \c buf
- * \param[in] user_buf (optional) buffer containing bytes allocated with
- * sufficient size for \c T. If this buffer is passed, the caller is responsible
+ * \param[in] object_buf (optional) buffer containing bytes allocated with
+ * sufficient size for \c T. If this buffer != null, the caller is responsible
```
PhilMiller, line 99: I think it should now be asserted that for this function, `object_buf != nullptr`, and this should never allocate.

The other overload can trivially be implemented in terms of this one.

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -155,19 +154,16 @@ buffer::ImplReturnType serializeType(T& to_serialize, BufferObtainFnType fn) {
...
 
 template <typename T>
-T* deserializeType(
-  SerialByteType* data, SerialSizeType const& size, T* allocBuf
-) {
-  auto mem = allocBuf ?
-    reinterpret_cast<SerialByteType*>(allocBuf) : new SerialByteType[sizeof(T)];
-  auto& t = Dispatch<T>::unpackType(mem, data, size);
+T* deserializeType(SerialByteType* data, SerialByteType* allocBuf) {
+  auto mem = allocBuf ? allocBuf : new SerialByteType[sizeof(T)];
```
PhilMiller, line 158: Seems pedantic, but maybe important in the long run (i.e. once we're leaning on C++17 support): this needs to be `new (std::align_val_t{alignof T}) SerialByteType[sizeof(T)]`

PR #92, opened 2020-05-06
91 Basic IO buffer with mmap

### LINKED ISSUE #88: Implement scalable disk IO for serialize-to-disk
Implement, in checkpoint, a scalable API for reading and writing objects to disk after serialization.

Phil's  comment:

> I realized earlier today that the right implementation might be really easy to bolt into the memory buffer model.

> - Do the sizing as usual
> - Open a file
> - (if packing) fallocate that file to the right size
> - mmap the file
> - pass the address to the normal memory buffer serializer
>  - (optionally call madvise(MADV_WONTNEED) every few megabytes)

> This lets the kernel and filesystem take care of things, and somewhat naturally paces operation to the filesystem's throughput without really blocking anywhere.

### LINKED ISSUE #91: Add custom traversals for serialize
Allow custom traversals other than size/pack/unpack for traversing data through serialize method.

lifflander:
Fixes #91 

Implements a new `IOBuffer` for writing and reading to a file by using mmap.

TODO:
- [x] Implement output buffer for `serializeToFile(..)`
- [x] Implement input buffer for `deserializeFromFile(..)`
- [x] Write tests for output/input to file


src/checkpoint/buffer/io_buffer.cc
```diff
@@ -0,0 +1,106 @@
...
+void IOBuffer::setupFile() {
+  debug_checkpoint("IOBuffer: opening file: %s\n", file_.c_str());
+
+  fd_ = open(file_.c_str(), O_CREAT | O_RDWR | O_TRUNC, (mode_t)0600);
+  assert(fd_ != -1 && "open must be valid");
+
+  debug_checkpoint("IOBuffer: truncating file\n");
+
+  int ret = ftruncate(fd_, size_);
+  assert(ret == 0 && "ftruncate should not fail");
```
PhilMiller: These will need to be actual runtime checks, not assertions. They can and do fail.

In these cases, I'd recommend throwing an exception. During checkpoint, the top-level code will have to decide whether failure to write is fatal or not, or should provoke a retry. During restart, it will obviously always be fatal.

src/checkpoint/buffer/io_buffer.cc
```diff
@@ -0,0 +1,106 @@
...
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include <assert.h>
+
+namespace checkpoint { namespace buffer {
+
+void IOBuffer::setupFile() {
+  debug_checkpoint("IOBuffer: opening file: %s\n", file_.c_str());
```
PhilMiller: There should be a check here that if we're somehow in the weird case that we don't have the 64-bit system calls, `size_` doesn't exceed our capability. Again, runtime error, not an assertion.

src/checkpoint/buffer/io_buffer.cc
```diff
@@ -59,8 +59,73 @@
...
+
+  int ret = 0;
+
+  /*
+   * Start by opening the file, read-only mode
+   */
+  fd_ = open(file_.c_str(), O_RDONLY, (mode_t)0600);
+  if (fd_ == -1) {
+    auto err = std::string("Failed to open file=") + file_ + ": errno=" +
+      std::to_string(errno);
```
PhilMiller: You could include `perror(errno)` in these strings, so that one doesn't need to look up meanings.

src/checkpoint/buffer/io_buffer.cc
```diff
@@ -117,20 +182,20 @@ void IOBuffer::setupFile() {
...
+  addr = mmap64(nullptr, size_, PROT_WRITE, MAP_SHARED, fd_, 0);
 
   if (addr == MAP_FAILED) {
     auto err = std::string("mmap64 failed for writing file: errno=") +
                std::to_string(errno);
     throw std::runtime_error(err);
   }
 # else
-  addr = mmap(nullptr, size_, PROT_WRITE, MAP_PRIVATE, fd_, 0);
+  addr = mmap(nullptr, size_, PROT_WRITE, MAP_SHARED, fd_, 0);
```
PhilMiller, line 199: I'm a little confused how `MAP_PRIVATE` was a defect in the write path - when should multiple processes have a shared mapping of a file that should be unique per process?

lifflander: Yes, it was writing a all-zero file with `MAP_PRIVATE`. Not sure why.

PhilMiller (approved):
All minor suggestions. Address as you see fit, and go ahead with merge

src/checkpoint/buffer/io_buffer.cc
lifflander: Ok, I've added `strerror(errno)` to the runtime error (can't use perror because it just prints to stderr, doesn't append to my error mesage).

src/checkpoint/buffer/io_buffer.cc
lifflander: BTW, Darwin doesn't have mmap64!

lifflander: I don't know how to detect easily if the signature of mmap (not mmap64) is 64-bit or not. I'm going to leave this until later.

lifflander: Actually, I'm wondering if I should detect and call `mmap2` in that case!

PR #94, opened 2020-05-07
93 General traversals

### LINKED ISSUE #93: Cleanup on traversal and dispatch
Refactor the dispatch code for general traversals.

lifflander:
Fixes #93 

Implements general traversal and dispatching of serialize functions. Gives the user far more power to do custom things with serializers.

Here's a very simple example with a custom traverser:

```C++
struct TestObject {
  template <typename SerializerT>
  void serialize(SerializerT& s) {
    s | a | b;
  }
  int a;
  std::vector<int> b;
};

/// Custom traverser for printing raw bytes
struct PrintBytesTraverse : checkpoint::Serializer {
  PrintBytesTraverse() : checkpoint::Serializer(checkpoint::eSerializationMode::None) { }

  void contiguousBytes(void* ptr, std::size_t size, std::size_t num_elms) {
    printf("PrintBytesTraverse: size=%lu, num_elms=%lu\n", size, num_elms);
  }
};

int main(int, char**) {
  TestObject obj;

  // Traverse my_obj with a custom traverser that prints the bytes
  checkpoint::dispatch::Traverse::with<TestObject, PrintBytesTraverse>(obj);

  return 0;
}
```

Here's an example with a custom dispatcher. You could use this to do some custom operation for certain types, like RDMA for std::vector or Kokkos::View

```C++

/// Custom dispatcher to customizing type-specific traversal behavior
/// For non-vectors, dispatch normally. For vectors, print the data inside
struct CustomDispatch {
  static void serializeIntrusive(SerializerT& s, T& t) {
    t.serialize(s);
  }
  static void serializeNonIntrusive(SerializerT& s, T& t) {
    serialize(s, t);
  }
};

// std::vector specialization for dispatcher, vector is always non-intrusive so
// skip that overload
template <typename SerializerT, typename U>
struct CustomDispatch<SerializerT, std::vector<U>> {
  static void serializeNonIntrusive(SerializerT& s, std::vector<U>& t) {
    printf("Traversing vector: size=%lu\n", t.size());
    for (std::size_t i = 0; i < t.size(); i++) {
      printf("\t vector[%lu]=%s", i, std::to_string(t[i]).c_str());
    }
  }
};

int main(int, char**) {
  TestObject obj;

  // Traverse my_obj with a custom traverser that prints the bytes
  checkpoint::dispatch::Traverse::with<TestObject, PrintBytesTraverse, CustomDispatch>(obj);

  return 0;
}
```

### DIFF
--- a/examples/checkpoint_traversal.cc
+++ b/examples/checkpoint_traversal.cc
@@ -82,7 +82,7 @@ struct PrintBytesTraverse : checkpoint::Serializer {
   PrintBytesTraverse() : checkpoint::Serializer(checkpoint::eSerializationMode::None) { }
 
   void contiguousBytes(void* ptr, std::size_t size, std::size_t num_elms) {
-    printf("PrintBytesTraverse: size=%lu, num_elms=%lu\n", size, num_elms);
+    printf("PrintBytesTraverse: size=%zu, num_elms=%zu\n", size, num_elms);
   }
 };
 
@@ -92,7 +92,7 @@ struct TypedTraverse : checkpoint::Serializer {
 
   template <typename SerializerT, typename T>
   void contiguousTyped(SerializerT&, T*, std::size_t num_elms) {
-    printf("TypedTraverse: type is %s, num=%lu\n", typeid(T).name(), num_elms);
+    printf("TypedTraverse: type is %s, num=%zu\n", typeid(T).name(), num_elms);
   }
 };
 
--- a/examples/checkpoint_traversal.cc
+++ b/examples/checkpoint_traversal.cc
@@ -1,0 +1,139 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                           checkpoint_traversal.cc
+//                           DARMA Toolkit v. 1.0.0
+//                 DARMA/checkpoint => Serialization Library
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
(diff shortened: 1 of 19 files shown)


PhilMiller:
This would also make a '`write()` to file' traversal easy to implement, yes?

examples/checkpoint_traversal.cc
```diff
@@ -0,0 +1,139 @@
...
+  TestObject my_obj(TestObject::MakeTag{});
+
+  // Traverse my_obj with a custom traverser that prints the bytes
+  checkpoint::dispatch::Traverse::with<TestObject, PrintBytesTraverse>(my_obj);
+
+  // Traverse my_obj with a custom traverser that prints the types and lens
+  checkpoint::dispatch::Traverse::with<TestObject, TypedTraverse>(my_obj);
+
+  // Traverse my_obj with a custom dispatcher
+  checkpoint::dispatch::Traverse::with<TestObject, TypedTraverse, CustomDispatch>(my_obj);
```
PhilMiller: This is written as a smoke test, without verification of the expected results. Maybe use a traverser that matches up a count of elements in vectors or something?

PhilMiller (approved):
The test could use improvement, but I'm happy to see this go ahead

lifflander:
> The test could use improvement, but I'm happy to see this go ahead

I need to write a real test for this.. Will do.

PhilMiller (approved):

lifflander:
There's still an issue with virtual serialize to work out here. But, I am merging for now to more forward on beta.8.

PR #100, opened 2020-05-14
99 Add serializer for std::unique_ptr

### LINKED ISSUE #99: Implement serializers for std::unique_ptr<T>

lifflander:
Fixes #99

### DIFF
--- a/src/checkpoint/container/unique_ptr_serialize.h
+++ b/src/checkpoint/container/unique_ptr_serialize.h
@@ -57,7 +57,7 @@ void serialize(Serializer& s, std::unique_ptr<T>& ptr) {
 
   if (not is_null) {
     if (s.isUnpacking()) {
-      auto t = std::allocator<T>{}.allocate(1);
+      auto t = dispatch::Standard::allocate<T>();
       ptr = std::unique_ptr<T>(dispatch::Reconstructor<T>::construct(t));
     }
     s | *ptr;
--- a/src/checkpoint/container/unique_ptr_serialize.h
+++ b/src/checkpoint/container/unique_ptr_serialize.h
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                 checkpoint.h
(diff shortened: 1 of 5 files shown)


PhilMiller:
I could swear we had this in there already. Maybe in the wrong file or something?

PhilMiller:
OK, I was thinking of the identical logic in `serializeRCP` I guess

src/checkpoint/container/unique_ptr_serialize.h
```diff
@@ -0,0 +1,69 @@
...
+namespace checkpoint {
+
+template <typename Serializer, typename T>
+void serialize(Serializer& s, std::unique_ptr<T>& ptr) {
+  bool is_null = ptr == nullptr;
+  s | is_null;
+
+  if (not is_null) {
+    if (s.isUnpacking()) {
+      auto t = std::allocator<T>{}.allocate(1);
```
PhilMiller: If we want to be *really* paranoid, we could add an assertion here that `alignof(T) <= __STDCPP_DEFAULT_NEW_ALIGNMENT__`

PhilMiller: `static_assert` would be nice, but I'm not sure if that would run into SFINAE issues in this circumstance

PhilMiller (approved):

src/checkpoint/container/unique_ptr_serialize.h
PhilMiller: I think this allocation and reconstruction logic is likely duplicative of something that should be in functions called by `dispatch::deserialize` already. Let's make sure we get it right there, and then just call the same thing.

PhilMiller (changes requested):

lifflander:
@PhilMiller I've abstracted out allocation into `dispatch::Standard::allocate<T>`. I can't combine this with reconstruction easily. And we can't call into the API that returns a unique_ptr because that's only on the unpacking side.

PR #101, opened 2020-05-14
98 Serialization sanitizer support

### LINKED ISSUE #98: Implement runtime API traversal for clang serializer checking
Integrate with generated code for https://github.com/DARMA-tasking/checkpoint-member-analyzer

lifflander:
Fixes #98 

Initial implémentation of stack-based checker using the clang tooling generated checks from:

https://github.com/DARMA-tasking/serialization-sanitizer


src/checkpoint/dispatch/checker.h
```diff
@@ -0,0 +1,162 @@
...
+  void addSkip(void* addr) {
+    skips.insert(addr);
+  }
+
+  std::string getName() const { return name_; }
+
+private:
+  std::string name_ = 0;
+public:
+  std::unordered_set<void*> exists;
```
PhilMiller: Please document what each of these members contain.

src/checkpoint/dispatch/checker.h
```diff
@@ -0,0 +1,162 @@
...
+template <typename SerializerT, typename T>
+struct CounterDispatch {
+  static void serializeIntrusive(SerializerT& s, T& t) {
+    if (stack.size() > 0) {
+      stack.top().add(reinterpret_cast<void*>(&t));
+    }
+
+    stack.push(CountRecord{typeid(T).name()});
+    t.serialize(s);
+
```
PhilMiller: Please grab `stack.top()` into a local variable, and use it below.

src/checkpoint/dispatch/checker.h
```diff
@@ -0,0 +1,162 @@
...
+    for (auto&& check : stack.top().checks) {
+      auto addr = std::get<0>(check);
+      auto name = std::get<1>(check);
+      auto skip_iter = stack.top().skips.find(addr);
+      if (skip_iter == stack.top().skips.end()) {
```
PhilMiller: Could you write this as `if (skips.contain(addr)) continue;`?

src/checkpoint/dispatch/checker.h
```diff
@@ -0,0 +1,162 @@
...
+        auto exists_iter = stack.top().exists.find(addr);
+        if (exists_iter == stack.top().exists.end()) {
```
PhilMiller: Going by this, I think the fields of `CountRecord` should be named something like `all_members, ignored_members, serialized_members`

src/checkpoint/dispatch/checker.h
```diff
@@ -0,0 +1,162 @@
...
+    }
+    stack.push(CountRecord{typeid(T).name()});
+    //printf("\t (%zu) \t NI: starting: %s\n", st.size(), typeid(T).name());
+    serialize(s, t);
+    //printf("\t (%zu) \t NI: finishing: %s\n", st.size(), typeid(T).name());
+    stack.pop();
+  }
+};
+
+/// Custom traverser for printing raw bytes
```
PhilMiller: This comment was clearly copied and pasted without correction.

PhilMiller (changes requested):
As written, does whatever 'top-level' object that's rooting a serialization call get properly checked?  I'm not quite sure on the interplay of the serializer, the dispatcher, and the push and pop calls.

PhilMiller:
Also, I looked over the corresponding bits in the clang tool, and they seem reasonable.

PhilMiller:
There's a potential corner case of objects with empty-base optimizations in play, but that's so obscure as to likely be irrelevant.

PhilMiller:
This side of things all looks good to me.

lifflander:
@PhilMiller @cz4rs Can you guys take a look at this? I've done a major overhaul of this code. I'm splitting up the concerns between the type-specific traversal for sanitization and the sanitizer runtime logic.

Checkpoint will contain the typed overloads at all times. This means one may use a stock VT/Checkpoint and then enable the sanitizer later.

The sanitizer runtime itself (`checkpoint::sanitizer::runtime`) will be in the other repository: https://github.com/DARMA-tasking/serialization-sanitizer

lifflander:
This won't be merged until the runtime is complete in the other repo and I confirm this is the correct interface that we need for that work properly.

src/checkpoint/serializers/sanitizer.h
```diff
@@ -0,0 +1,253 @@
...
+   * \param[in] tinfo the name of the type recursed out of
+   */
+  virtual void pop(std::string tinfo) {}
+
+};
+
+/// pimpl to runtime that contains runtime sanitizer logic
+extern Runtime* rt();
+
+/// function that informs sanitizer if its enabled
```
PhilMiller, line 115: Typo "it's"

src/checkpoint/serializers/sanitizer.impl.h
```diff
@@ -0,0 +1,139 @@
...
+}
+
+template <typename SerT, typename T>
+void SanitizerDispatch<SerT,T>::serializeIntrusive(SerT& s, T& t) {
+  if (not sanitizer::enabled()) {
+    return;
+  }
+
+  // Declare this member as serialized
+  sanitizer::rt()->isSerialized(cleanTypeToVoid(t), 1, typeid(T).name());
```
PhilMiller: Could we have the runtime be a dependency injection at construction, rather than accessed at need through a static function? It would make the coupling clearer, and clean up potential tests.

src/checkpoint/serializers/sanitizer.impl.h
```diff
@@ -0,0 +1,139 @@
...
+
+  // Push this type on the stack since we are about to traverse it
+  sanitizer::rt()->push(typeid(T).name());
+
+  // Recurse with the current serializer (a sanitizing pass)
+  t.serialize(s);
+
+  // A bit of a trick...recurse with a non-sanitizing-specific overload to
+  // resolve actual \c serialize methods without invoking the partial
+  // specialization
```
PhilMiller, line 79: I think this comment may need more elaboration. I think I kinda get what's going on, but would prefer a full explanation.

src/checkpoint/serializers/sanitizer.cc
```diff
@@ -0,0 +1,68 @@
...
+  static std::unique_ptr<sanitizer::Runtime> base_rt;
+  if (base_rt == nullptr) {
+    base_rt = std::make_unique<sanitizer::Runtime>();
+  }
+  return base_rt.get();
+}
+
+bool enabled() {
+  return false;
+}
```
PhilMiller, line 65: Would it make sense to have this be a member of the `Runtime` object?

PhilMiller:
Broadly, I think this looks very good. I'll be happy to see this tidied up and merged.

src/checkpoint/serializers/sanitizer.cc
lifflander: These have to be C functions because they are LD preloaded by the serializer runtime.

PhilMiller: Could you point me at where you picked up that constraint? That doesn't sound right to me, since `LD_PRELOAD` should just be importing whatever symbols it finds in a `foo.so` library into the list that `ld.so` will resolve against. If those are C function names or C++ mangled names shouldn't matter.

That said, this isn't a big deal.

src/checkpoint/serializers/sanitizer.impl.h
```diff
@@ -0,0 +1,149 @@
...
+}
+
+template <typename SerT, typename T>
+void SanitizerDispatch<SerT,T>::serializeIntrusive(SerT& s, T& t) {
+  if (not checkpoint_sanitizer_enabled()) {
+    return;
+  }
+
+  // Declare this member as serialized
+  checkpoint_sanitizer_rt()->isSerialized(
```
PhilMiller, line 67: I think it would be preferable to do `auto rt = checkpoint_sanitizer_tr()` once here, and then just `rt->method()`

PhilMiller:
One important check - does this work if one *doesn't* include the `LD_PRELOAD`?

Otherwise, I think this is ready to move forward, modulo my remarks about some comments and style.

cz4rs (approved):
looks good overall, I agree with Phil on "trick" part needing more explanation

PR #106, opened 2020-05-27
95 Rewrite virtual serialize

### LINKED ISSUE #95: Virtual serialization is not compatible with custom traversals
Virtual serialization uses the enum `eSerializationMode` to cast the traverser type during virtual dispatch when `SerializerT` is lost. Losing this means that we won't have the right traversal as specified by the user in the `TraverserT`.

lifflander:
Fixes #95

TODO:
 - [x] Fix default instantiations
 - [x] Write tests for virtual serialize


src/checkpoint/dispatch/vrt/derived.h
```diff
@@ -0,0 +1,124 @@
...
+template <typename DerivedT, typename BaseT>
+struct SerializableDerived : BaseT {
+  template <typename... Args>
+  explicit SerializableDerived(Args&&... args)
+    : BaseT(std::forward<Args>(args)...)
+  { }
+
+  SerializableDerived() = default;
+
+  template <typename SerializerT>
```
lifflander: Do we actually need this?

src/checkpoint/dispatch/vrt/object_registry.h
```diff
@@ -59,7 +60,9 @@ template <typename T>
 using RegistryType = std::vector<
   std::tuple<
```
PhilMiller: Could we make these a vector of `struct` of some defined type, rather than a `tuple`, so that we don't need to keep track of what `get<2>` and such mean below?

src/checkpoint/dispatch/vrt/virtual_serialize.h
```diff
@@ -0,0 +1,158 @@
...
+template <typename T, typename SerializerT>
+void serializeAllocationWrapper(T*& object, SerializerT& s) {
+  TypeIdx entry = -1;
+
+  if (not s.isUnpacking()) {
+    entry = RegisterIfNeeded<T>::apply(object);
+  }
+
+  OptionalTypeToken token(VirtualSerializeTraits<T>::has_virtual_serialize, entry);
+  s | token;
```
PhilMiller: If I understand correctly, it looks like we serialize one of these for each `unique_ptr` object serialized. I'm not necessarily objecting to that as unreasonable overhead, but I think we could avoid it statically for the non-virtual cases.

src/checkpoint/dispatch/vrt/base.h
```diff
@@ -68,16 +70,11 @@ struct SerializableBase : SerializableBaseBase {
...
 
-    if (base_idx != expected_idx) {
-      printf("broken assert %s\n", typeid(BaseT).name());
-    }
-
-    checkpointAssert(
-      base_idx == expected_idx,
-      "Type idx for base does not matched expected value. "
+    auto debug_str = std::string("Type idx for base \"") + typeid(BaseT).name() +
+      "\" does not matched expected value. "
```
PhilMiller: This string is going to be constructed for every object that goes through serialization, every time it does so.

It either needs to be formed as part of debug-guarded execution, or maybe just made `static` so that it only happens once per type.

src/checkpoint/dispatch/vrt/derived.h
```diff
@@ -86,17 +88,17 @@ struct SerializableDerived : BaseT {
...
-      }
-      checkpointAssert(
-        derived_idx == expected_idx or expected_idx == -1,
-        "Type idx for derived class does not matched expected value. "
-        "You are probably missing a SerializableBase<T> or SerializableDerived<T> "
-        "in the virtual class hierarchy."
-      );
-    }
+
+    auto debug_str = std::string("Type idx for derived \"") +
```
PhilMiller: Ditto string formation

src/checkpoint/traits/checkpoint_traits.h
```diff
@@ -77,10 +77,22 @@ struct SerializerTraits {
...
 
+  template <typename U>
+  using isVirtualDisabled_t = decltype(std::declval<const U>().isVirtualDisabled());
+  using has_isVirtualDisabled = detection::is_detected_convertible<bool, isVirtualDisabled_t, T>;
+
+  template <typename U>
+  using setVirtualDisabled_t = decltype(std::declval<U>().setVirtualDisabled(std::declval<bool>()));
+  using has_setVirtualDisabled = detection::is_detected_convertible<bool, setVirtualDisabled_t, T>;
+
   // This defines what it means to be reconstructible
```
PhilMiller: This comment is misplaced

src/checkpoint/container/unique_ptr_serialize.h
```diff
@@ -58,7 +58,7 @@ void serialize(Serializer& s, std::unique_ptr<T>& ptr) {
 
   if (not is_null) {
     T* t = ptr.get();
-    dispatch::vrt::serializeAllocatePointer<T>(t, s);
+    serializeAllocatePointer(s, t);
```
PhilMiller: Just a naming nitpick: it's not allocating a pointer, it's allocating (space for) an object

src/checkpoint/dispatch/vrt/base_base.h
```diff
@@ -52,15 +52,7 @@ namespace checkpoint { namespace dispatch { namespace vrt {
...
-struct SerializableBaseBase {
-
-  virtual void _checkpointDynamicSerialize(
-    void* s, TypeIdx serializer_idx, TypeIdx expected_idx
-  ) = 0;
-
-  virtual TypeIdx _checkpointDynamicTypeIndex() = 0;
-
-};
+struct SerializableBaseBase { };
```
PhilMiller: I think the relevant traits don't look at this anymore, checking detection of the necessary methods directly?

src/checkpoint/dispatch/reconstructor.h
```diff
@@ -181,6 +189,26 @@ struct Reconstructor {
...
+  /// Overloads that allow failure to reconstruct so SFINAE overloads don't
+  /// static assert out
+  template <typename U = T>
+  static T* constructAllowFailImpl(void* buf, isConstructible<U>* = nullptr) {
+    return construct<U>(buf);
+  }
+
+  template <typename U = T>
+  static T* constructAllowFailImpl(void* buf, isNotConstructible<U>* = nullptr) {
+    checkpointAssert(false, "This code should be unreachable");
```
PhilMiller: This failure can be more specific - "Checkpoint is wrongly trying to reconstruct an abstract base class instance". At least, I think that's the case we're addressing. A derived type that just doesn't satisfy any of the reconstruction options will fail to compile, right?

src/checkpoint/dispatch/vrt/inheritance_assert_helpers.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace checkpoint { namespace dispatch { namespace vrt {
+
+template <typename ObjT>
+inline void assertTypeIdxMatch(TypeIdx const expected_idx) {
+  auto obj_idx = objregistry::makeObjIdx<ObjT>();
+
+  auto debug_str = std::string("Type idx for object \"") + typeid(ObjT).name() +
+    "\" does not matched expected value. "
+    "You are probably missing a SerializableBase<T> or SerializableDerived<T> "
```
PhilMiller, line 62: should this reference the alternative macros as well?

src/checkpoint/serializers/serializers_headers.h
```diff
@@ -51,4 +51,12 @@
 #include "checkpoint/serializers/packer.h"
 #include "checkpoint/serializers/unpacker.h"
 
+#define checkpoint_serializer_variatic_args()   \
```
PhilMiller: Typo: variadic

src/checkpoint/dispatch/vrt/serializer_registry.h
```diff
@@ -109,18 +116,33 @@ TypeIdx const Type<ObjT, SerializerT>::idx =
...
-inline TypeIdx makeObjIdx() {
-  return Type<ObjT, SerializerT>::idx;
+template <typename ObjT>
+inline auto getObjIdx(TypeIdx han) {
+  auto elm = getObjIdxRef<ObjT>(han);
+  return std::get<2>(elm);
+}
+
+template <typename DerivedT>
+inline std::function<void(void*, DerivedT&)> getBaseIdx(TypeIdx base_idx) {
```
PhilMiller, line 149: I don't see this actually being used.

If it is, I don't understand at all what it's doing.

src/checkpoint/dispatch/vrt/serialize_instantiator.h
```diff
@@ -0,0 +1,92 @@
...
+#include "checkpoint/dispatch/vrt/link_derived_to_base_idx.h"
+
+namespace checkpoint { namespace dispatch { namespace vrt {
+
+template <typename ObjT, typename SerializerT>
+static inline void instantiateObjSerializerImpl() {
+  dispatch::vrt::serializer_registry::makeObjIdx<ObjT, SerializerT>();
+
+  // We must save the idx of base in derived so we ensure we get the right entry
+  // if they are unaligned
```
PhilMiller: What does 'unaligned' mean here? It's clearly not about alignment in the plain sense.

src/checkpoint/dispatch/vrt/serializer_registry.h
```diff
@@ -0,0 +1,150 @@
...
+  );
+
+  reg.emplace_back(
+    index,
+    no_type_idx, // Set later when linkDerivedToBase is called
+    [=](void* s, ObjT& obj) {
+      auto& ser = *reinterpret_cast<SerializerT*>(s);
+      // Disable virtual serializer dispatch because we are already in a
+      // virtualSerialize and otherwise we will recurse indefinitely
+      ser.setVirtualDisabled(true);
```
PhilMiller: Does this need to be a stack where we save the value before pushing 'true' here, and restore it when we come back out? Consider recursive nested objects.

src/checkpoint/dispatch/vrt/derived.h
PhilMiller: I think not - add an assertion and check that it's never called. It shouldn't be, I don't think.

src/checkpoint/dispatch/vrt/serializer_registry.h
PhilMiller: It see it's used in a later commit. I still don't really understand what it's doing.

PhilMiller:
Could you either find a suitable place for overview comments, or write a 'cover letter' for the overall PR in the description, that explains the overall working of the pieces here?

src/checkpoint/dispatch/vrt/serializer_registry.h
lifflander: It gets immediately unset when it triggers the non-virtual serialize in the dispatcher. I have a test with nesting to ensure this works properly.

src/checkpoint/traits/serializable_traits.h
```diff
@@ -117,6 +118,10 @@ struct SerializableTraits {
   using constructor_t = decltype(U());
   using has_default_constructor = detection::is_detected<constructor_t, T>;
 
+  template <typename U>
+  using tagged_constructor_t = decltype(U(std::declval<SERIALIZE_CONSTRUCT_TAG>()));
+  using has_tagged_constructor = detection::is_detected<tagged_constructor_t, T>;
+
   template <typename U>
   using reconstruct_t = decltype(U::reconstruct(std::declval<void*>()));
   using has_reconstruct =
```
PhilMiller: I think this needs to be stricter, to avoid the case where a parent class (direct or not) has a `reconstruct` member, but `U` does not.

One consideration is that we need `is_same<T, reconstruct_t>` rather than convertibility - otherwise, we're calling the wrong serializer.

Another consideration is something along the lines of 
```
no_base_type or not has_reconstruct<BaseType> or &static_cast<U::*(void*)>(U::reconstruct) != &static_cast<U::*(void*)>(Base::reconstruct)
```
I.e. that if this is a derived class, and some base defines a `reconstruct` method, that's not the one we're going to try to call.

src/checkpoint/dispatch/vrt/base_base.h
PhilMiller: I.e. I think we can just remove this.

src/checkpoint/dispatch/vrt/serialize_instantiator.h
lifflander: If the `SerializerT` static registrations order happen differently for a given `BaseT` and `DerviedT` (which I'm calling unaligned), you will get the wrong serializer when using the base_serializer_idx in the derived class. This was an issue I imagined before, and it actually did happen when I starting running larger examples. Thus, I link the correct base serialize type idx with a given derived type idx during sizing/packing. Then, when it unpacks, it can find the right one when running on the derived type using the base idx.

src/checkpoint/dispatch/vrt/base.h
```diff
@@ -0,0 +1,94 @@
...
+    ::checkpoint::dispatch::vrt::TypeIdx expected_idx                             \
+  ) {                                                                             \
+    ::checkpoint::dispatch::vrt::assertTypeIdxMatch<BASE>(expected_idx);          \
+    auto dispatcher =                                                             \
+      ::checkpoint::dispatch::vrt::serializer_registry::getObjIdx<BASE>(ser_idx); \
+    dispatcher(s, *static_cast<BASE*>(this));                                     \
+  }                                                                               \
+  virtual ::checkpoint::dispatch::vrt::TypeIdx _checkpointDynamicTypeIndex() {    \
+    return ::checkpoint::dispatch::vrt::DispatchTypeIdx<BASE>::get();             \
+  }                                                                               \
```
PhilMiller, line 72: Do we need the instantiation call in this macro as well? I think it may only complain in a case where a program defines a non-pure base, and no derived classes that would end up instantiating it by recursion.

src/checkpoint/dispatch/vrt/serialize_instantiator.h
lifflander: Note, the linking happens at runtime after the registrations are done and the type indices are set.

src/checkpoint/dispatch/vrt/base.h
lifflander: Good point.

src/checkpoint/dispatch/dispatch_serializer_nonbyte.h
```diff
@@ -188,6 +198,25 @@ struct SerializerDispatchNonByte {
       applyElm(s, val+i);
     }
   }
+
+  template <typename U = T>
+  void apply(
+    SerializerT& s, T* val, SerialSizeType num, hasVirtualSerialize<U>* = nullptr
+  ) {
+    if (s.isVirtualDisabled()) {
+      s.setVirtualDisabled(false);
```
PhilMiller, line 207: My remark about keeping a stack of virtual dispatch states would potentially apply here as well.

lifflander: I don't think we need a stack. It needs to be set for one layer and then unset immediately.

src/checkpoint/dispatch/vrt/virtual_serialize.h
lifflander: You are correct. I implemented it this way out of simplicity. I could do some SFINAE on the type to get rid of that bool. Not sure it's worth the effort.

src/checkpoint/dispatch/vrt/serializer_registry.h
lifflander: Initially, I thought a stack might be needed also. But I believe a stack isn't needed because the state is only valid for exactly one call depth and then unset.

src/checkpoint/container/unique_ptr_serialize.h
lifflander: What would you call this? I imagine users calling this for raw pointers. I was thinking serializePointer, but I want to ensure it's clear that this is going to do allocation.

src/checkpoint/dispatch/vrt/serialize_instantiator.h
PhilMiller: Ok. Could you rewrite the comment to describe that?

src/checkpoint/dispatch/vrt/derived.h
```diff
@@ -0,0 +1,124 @@
...
+      DERIVED,                                                                       \
+      checkpoint_serializer_variatic_args()                                          \
+    >();                                                                             \
+    debug_checkpoint(                                                                \
+      "%s: BEGIN: _checkpointDynamicSerialize: serializer_idx=%d {\n",               \
+      typeid(DERIVED).name(), base_ser_idx                                           \
+    );                                                                               \
+    ::checkpoint::dispatch::vrt::assertTypeIdxMatch<DERIVED>(expected_idx);          \
+    auto base_idx = ::checkpoint::dispatch::vrt::objregistry::makeObjIdx<BASE>();    \
+    BASE::_checkpointDynamicSerialize(s, base_ser_idx, base_idx);                    \
```
lifflander, line 70: Do I need some extra disambiguation here if BASE is actually a more complex type with template parameters? What happens when it's not?

src/checkpoint/dispatch/vrt/static_dispatch_typeidx.h
```diff
@@ -0,0 +1,81 @@
...
+#include <type_traits>
+
+namespace checkpoint { namespace dispatch { namespace vrt {
+
+template <typename T, typename _enabled = void>
+struct DispatchTypeIdx;
+
+template <typename T>
+struct DispatchTypeIdx<
+  T, typename std::enable_if_t<std::is_abstract<T>::value>
```
PhilMiller: Maybe I'm not understanding, but the cases here feel backwards to me:
- `is_abstract => no_type_idx`
- `not is_abstract => makeObjIdx()`

src/checkpoint/dispatch/vrt/derived.h
PhilMiller: I think because you're ultimately calling a function, all should be well. I can see how we might get into a weird space where `typename` is hypothetically required of us for some uses, and forbidden for others.

Or is your concern that `DERIVED` and/or `BASE` might contain commas in template arguments, thus sans parentheses, and thereby confuse the preprocessor about how many arguments are being passed to the macro? We may need to require and document that arguments to these macros have to be either templates of just one argument, or a typedef to a single symbol, to avoid that problem.

src/checkpoint/dispatch/vrt/virtual_serialize.h
PhilMiller: If we implement any sort of debugging aid that prints out what's being serialized, this'll look embarrassing. I think it may be worth it for that, in addition to the concrete overhead.

src/checkpoint/dispatch/vrt/static_dispatch_typeidx.h
lifflander: I actually think `DispatchTypeIdx` can be removed entirely. With `constructAllowFail` in place, registering regardless should be fine (with the caveat that we might have some extra registrations that own't hurt). We still want dynamic dispatch to get the top level object idx, but we don't need to not register if it's abstract. That should simplify the code. What do you think?

src/checkpoint/container/unique_ptr_serialize.h
PhilMiller: It's already in the namespace, so just `allocate`? That's what it was previously, as `dispatch::Standard::allocate`, and I thought that was fine.

src/checkpoint/dispatch/vrt/derived.h
lifflander: I was thinking about the typename issue, but I think the commas are potentially an issue also. Requiring a typedef in those cases seems reasonable. Extra parens will make C++ unhappy here.

A typedef can also solve the potential typename ambiguity issue also.

src/checkpoint/dispatch/vrt/static_dispatch_typeidx.h
PhilMiller: Anything that can simplify out any of this machinery at all would be *very* welcome.

src/checkpoint/dispatch/vrt/base.h
lifflander: Does that equate to using virtual serialize without any derived classes? The way I see it, if you have any derived classes at all you will recursively reference the base at some point. I guess it's a valid use mode to not have derived classes, maybe using virtual serialize because you plan to add some later or you plan for code extendability. In fact, this might happen with EMPIRE classes that get derived classes in optional components.

src/checkpoint/dispatch/vrt/derived.h
lifflander: On the checkpoint end, I should add comments about this *and* in `Serializable{Base,Derived}` make a typedef before calling the macro.

PhilMiller:
So, on an especially terrifying note, this could in theory be extended to handle diamond inheritance automatically.

src/checkpoint/dispatch/vrt/object_registry.h
lifflander: done

src/checkpoint/dispatch/vrt/virtual_serialize.h
lifflander: done

src/checkpoint/dispatch/vrt/base.h
lifflander: done

src/checkpoint/traits/checkpoint_traits.h
lifflander: done

src/checkpoint/container/unique_ptr_serialize.h
lifflander: Ok, this call is different because it actually allocates+constructs. Maybe `allocateConstructForPointer`

lifflander: done

src/checkpoint/dispatch/vrt/base_base.h
lifflander: done

src/checkpoint/dispatch/reconstructor.h
lifflander: That is the case it addresses. fixed.

src/checkpoint/serializers/serializers_headers.h
lifflander: Fixed

src/checkpoint/dispatch/vrt/serialize_instantiator.h
lifflander: done

src/checkpoint/dispatch/vrt/static_dispatch_typeidx.h
lifflander: removed

src/checkpoint/dispatch/vrt/base.h
lifflander: Added to base

PhilMiller:
So, the ugly question- does this blow up Intel or nvcc?

PhilMiller:
I realize that of all possible multiple inheritance cases, diamond inheritance is actually the easiest for us to address here, since it has a common base class!

src/checkpoint/traits/serializable_traits.h
lifflander: Turns out that the return type of reconstruct makes it distinct. Even with `is_detected_convertible` this checks for an **implicit** conversion from Base to Derived. The only way that could potentially be valid is with a constructor in Derived that takes Base as an argument. However, since this is checking convertibility on a reference to derived, it even that doesn't trigger it.

lifflander: I even tried getting it to fail with `operator BaseT&() & { return *this; }`, but this didn't trigger it either. Note, with the constructor implicit converter I got `std::is_convertible<BaseT, DerviedT>::value` to be true, but not `std::is_convertible<BaseT&, DerviedT&>::value`

lifflander: I've changed this to `is_detected_exact` to be super safe and added a test to ensure it works in this case.

lifflander:
> I realize that of all possible multiple inheritance cases, diamond inheritance is actually the easiest for us to address here, since it has a common base class!

I was thinking the same thing. We could detect and address that one pretty easily and cleanly do the check to make sure that the base is only serialized once.

lifflander:
> So, the ugly question- does this blow up Intel or nvcc?

I'm trying intel right now. nvcc is another nightmare... Really for nvcc, we need to check in in conjunction with EMPIRE since that combo might be problematic.

lifflander:
@PhilMiller Do you think it's worth adding diamond here? That extension would address a big use case in EMPIRE right now.

Also, I think I've addressed all your concerns. Can you ensure I did to your satisfaction?

PhilMiller:
I'd hold off on diamond, definitely in this PR, and possibly until we have a second expected use case

PhilMiller:
If we do address 'diamond' inheritance, we'll need to be very clear that we really mean any semi-lattice of types - they must all share a common ultimate ancestor

src/checkpoint/dispatch/vrt/object_registry.h
```diff
@@ -93,12 +107,12 @@ Registrar<ObjT>::Registrar() {
   debug_checkpoint("object registrar: %zu, %s\n", reg.size(), typeid(ObjT).name());
 
   reg.emplace_back(
-    std::make_tuple(
+    ObjectEntry<BaseType>{
       index,
       []()          -> std::size_t { return sizeof(ObjT); },
```
PhilMiller: Couldn't this just be `size_t`, evaluated right here?

src/checkpoint/dispatch/vrt/virtual_serialize.h
```diff
@@ -132,72 +66,67 @@ void virtualSerialize(T*& base, SerializerT& s) {
...
       // use type idx here, registration needed for proper type re-construction
-      auto idx = token.getTypeIdx();
-      auto t = dispatch::vrt::objregistry::getObjAllocate<T>(idx)();
-      target = dispatch::vrt::objregistry::getObjConstruct<T>(idx)(t);
-    } else {
-      // no type idx needed in this case, static construction in default case
-      // we can't directly call because of bad mix between SFINAE and static_assert
-      target = StaticAllocateConstruct<T>::apply();
+      auto t = dispatch::vrt::objregistry::getObjAllocate<T>(entry)();
+      target = dispatch::vrt::objregistry::getObjConstruct<T>(entry)(t);
```
PhilMiller: Looking at this, might a slight refactoring to just give the registry functions the call the registered function objects and return their results be cleaner?

I.e. 
```
auto t = dispatch::vrt::objregistry::allocate<T>(entry);
target = dispatch::vrt::objregistry::construct<T>(entry, t);
```

src/checkpoint/dispatch/vrt/object_registry.h
lifflander: Yep, no reason for that to be a lambda. Currently, this overload isn't even called, but I'm imagining a future variant where the user gets to control allocation

src/checkpoint/dispatch/vrt/virtual_serialize.h
lifflander: Sure, why not!

src/checkpoint/dispatch/dispatch_virtual.h
```diff
@@ -48,4 +48,115 @@
...
+ *     - \c checkpoint_virtual_serialize_dervied(T,BaseT)
+ *
+ * Invoking the virtual serializer:
+ *
+ *   - If you have a \c std::unique_ptr<T>, where T is virtually serializable (by
+ *     using the macros or inheriting from \c SerializableBase<T> and
+ *     \c SerializableDervived<T,BaseT> ), they will automatically be virtually
+ *     serialized.
+ *
+ *   - If you have a raw pointer, \c Teuchos::RCP, or \c std::shared_ptr<T>,
```
PhilMiller, line 81: I would word this as "If you own the object reference by a raw pointer, ..."

src/checkpoint/dispatch/vrt/object_registry.h
lifflander: Done

src/checkpoint/dispatch/vrt/virtual_serialize.h
lifflander: Done

src/checkpoint/dispatch/dispatch_virtual.h
```diff
@@ -48,4 +48,115 @@
...
+ *           bool is_null = raw_ptr == nullptr;
+ *           s | is_null; // serialize whether we have a null pointer not
+ *           if (!is_null) {
+ *             // During size/pack, save the actual derived type of raw_ptr;
+ *             // During unpack, allocate/construct raw_ptr with correct virtual type
+ *             checkpoint::allocateConstructForPointer(s, raw_ptr);
+ *             s | *raw_ptr;
+ *           }
+ *         }
+ *       };
```
PhilMiller, line 101: Could we extract this motif into a function that any of these pointer holders can easily call?

```c++
void serializeOwningPointer(SerializerT& s, T*& raw_ptr) {
  bool is_null = raw_ptr == nullptr;
  s | is_null; // serialize whether we have a null pointer not
  if (!is_null) {
    // During size/pack, save the actual derived type of raw_ptr;
    // During unpack, allocate/construct raw_ptr with correct virtual type
    checkpoint::allocateConstructForPointer(s, raw_ptr);
    s | *raw_ptr;
  }
}
```

src/checkpoint/dispatch/dispatch_virtual.h
```diff
@@ -48,4 +48,115 @@
...
+ *
+ * ---------------------------------------------------------------------------
+ * --------------------- Overview of implementation: -------------------------
+ * ---------------------------------------------------------------------------
+ *
+ * When packing we need to traverse the whole hierarchy of virtual classes of
```
PhilMiller: Typo 'of'?

src/checkpoint/dispatch/dispatch_virtual.h
lifflander: Hmm, yes I think we could: In fact, we can provide some overloads on top of it for things like shared_ptr, or maybe even this:

```C++
template <typename SerializerT, template <typename T> class SmartPtrLike>
void serializeOwningPointer(SerializerT& s, SmartPtrLike<T>& ptr) {
  // call down
}
```

lifflander:
For some reason, all the checkpoint tests are failing/segfaulting on Mutrino due to some weird compilation problem probably some issue with gtest.

But, the examples do run correctly, including the virtual serialize one!

src/checkpoint/dispatch/dispatch_virtual.h
```diff
@@ -48,4 +48,115 @@
...
+ * pack of all existing serializers that ship with checkpoint.
+ *
+ * *Important*: if you define a custom traversal, you need to manually
+ * instantiate it for all virtual objects.
+ *
+ * When a serializer is looked up in the registry, it gets an index for a given
+ * \c T -- where \c T is a user's virtually serialized object type. However, we
+ * save the registry index for the base class in the hierarchy. Because the
+ * serializer registry is templated on \c T (i.e., there is a separate registry
+ * for each \c T ) we must match up the entry across registries for a derived
```
PhilMiller, line 139: Maybe I'm being dense or forgetful, but why can't we have a typedef in `SerializableBase<BaseT>`, `using _checkpointUltimateBaseType = BaseT`, and then always refer to registries directly by that ultimate base type?

I feel like something like that was there, and I can't tell if that was taken out because it doesn't work for some reason, or just because it got dropped on the floor.

lifflander: If I understand your remark correctly---I started with that, which in theory would have simplified things.

However, it creates undefined behavior because you have a dependency between two static initializations (the ordering issue). Basically, while you are statically initializing `DerviedT`, you try to generate the index for the base---static initialize `BaseT`.

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: I'd be really hesitant to write something that attempts to be generic across smart pointer types. It's a pretty broad field of possibilities, and we don't know what's going to break if we get it wrong.

I'd lean toward just making it very easy to implement explicit functions for specific smart pointer types, and providing this overload as suggested only for `shared_ptr`, and maybe with library availability checking for `Teuchos::RCP`

src/checkpoint/dispatch/dispatch_virtual.h
lifflander: Oh, if you're saying we should never register for the derived at all and just use the base, that's not possible because the operator| with SerializerT must be invoked on the *derived* type with that exact type.

src/checkpoint/traits/serializable_traits.h
```diff
@@ -171,9 +176,12 @@ struct SerializableTraits {
   static constexpr auto const has_serialize_function =
     has_serialize_instrusive or has_serialize_noninstrusive;
 
+  static constexpr auto const is_constructible =
+    is_default_constructible or is_reconstructible or is_tagged_constructible;
```
PhilMiller, line 206: This is missing the `is_nonintrusive_reconstructible` case I think

lifflander: Good point, I should rename `is_reconstructible` as `is_intrusive_reconstructible` to make that clear and create the OR-combo as `is_reconstructible`

src/checkpoint/traits/serializable_traits.h
```diff
@@ -171,9 +176,12 @@ struct SerializableTraits {
...
+
   // This defines what it means to be serializable
   static constexpr auto const is_serializable =
-    has_serialize_function and (is_default_constructible or is_reconstructible);
+    has_serialize_function and is_constructible;
```
PhilMiller, line 212: Could you order the definitions in this class to group them by 'category' of use? I.e. byte copy and arithmetic, then construction, then serialization, then composites that mix those?

I'm fine with having all the `constexpr` values after all the type definitions, as long as the grouping is applied within each. It's just jarring to see everything kinda tossed together.

src/checkpoint/dispatch/vrt/object_registry.h
```diff
@@ -132,12 +132,12 @@ inline auto getObjSize(TypeIdx han) {
 
 template <typename T>
 inline auto getObjAllocate(TypeIdx han) {
```
PhilMiller: These now ought not to be named `getObjFoo`

How about `allocateConcreteType` and `constructConcreteType`?

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: I agree, we clearly need to register every derived type, to actually invoke things properly.

I'm unclear about the UB that would arise from consolidating the registry. Do you mean true UB, or just that the order and hence indices would be unspecified? I think we might be able to impose some sort of natural ordering, such as the ultimate base always being index 0, and a topological ordering in which parents in turn precede children.

src/checkpoint/dispatch/vrt/object_registry.h
```diff
@@ -0,0 +1,150 @@
...
+TypeIdx const Type<ObjT>::idx = Registrar<ObjT>().index;
+
+template <typename T>
+inline auto getObjIdx(TypeIdx han) {
+  debug_checkpoint("getObjIdx: han=%d, size=%ld\n", han, getRegistry<T>().size());
+  return getRegistry<T>().at(han).idx_;
+}
+
+template <typename T>
+inline auto getObjSize(TypeIdx han) {
```
PhilMiller: I feel silly asking this now, but is `getObjSize` actually used anywhere? Can the size field go away entirely? Is there something we anticipate needing it for if not presently?

src/checkpoint/dispatch/vrt/serializer_registry.h
```diff
@@ -0,0 +1,150 @@
...
+
+namespace checkpoint { namespace dispatch { namespace vrt {
+namespace serializer_registry {
+
+template <typename ObjT>
+using RegistryType = std::vector<
+  std::tuple<
+    TypeIdx,                             /**< This entry type idx */
+    TypeIdx,                             /**< The base class type idx */
+    std::function<void(void*, ObjT&)>    /**< Type-erased serialize dispatch */
```
PhilMiller: As with the object registry, this would be better as a struct with named members.

src/checkpoint/dispatch/vrt/serializer_registry.h
PhilMiller: Ok, I understand that now. It would be helpful to clarify in the comment that it immediately gets unset on the other side of the fence.

PhilMiller: And probably have a matching comment where that happens, pointing here.

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: https://isocpp.org/wiki/faq/ctors#static-init-order makes me think that this should be doable. If there's some subtlety that I'm missing, it sounded like you maybe had examples in hand?

src/checkpoint/traits/serializable_traits.h
lifflander: Ok, just pushed a whole re-order and cleanup of traits.

src/checkpoint/dispatch/vrt/object_registry.h
lifflander: Yes, maybe you missed my earlier comment, but I anticipate needing it for the future when we want to give the user more control over allocation. Currently, I don't have an overload for that in `allocateConstructForPointer`, but it could be easily adding with a callback.

src/checkpoint/dispatch/reconstructor.h
```diff
@@ -58,52 +59,172 @@ template <typename T>
...
+    return nullptr;
+  }
+
   #if HAS_DETECTION_COMPONENT
+
+  /*
+   * Try to reconstruct with the following precedence:
+   *
+   *   - Tagged constructor: T(SERIALIZE_CONSTRUCT_TAG{})
+   *   - Reconstruct: T::reconstruct(buf) or reconstruct(t, buf)
```
PhilMiller, line 146: I think `reconstruct` ought to come ahead of a tagged constructor, but that doesn't need to be resolved here and now. I think we're never going to see the two of them in conflict. Maybe we should just detect and error if they're ever both available?

src/checkpoint/dispatch/dispatch_virtual.h
lifflander: The registry is already in a function:
```C++
template <typename ObjT>
inline RegistryType<ObjT>& getRegistry() {
  static RegistryType<ObjT> reg;
  return reg;
}
```

I'm not sure how to phase the issue exactly (maybe it's a dynamic initialization dependency), but getting the index from another registry during `Registrar<ObjT, SerializerT>::Registrar()`, which happens at load time is definitely causing UB, because I see it generating zeros sometimes or the correct value other times which is the typical outcome from doing something wrong during static init

src/checkpoint/dispatch/vrt/object_registry.h
lifflander: done

src/checkpoint/dispatch/vrt/serializer_registry.h
lifflander: Done

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: If you're not already into it, I should say that this doesn't need to be part of this PR. It's an unrelated convenience.

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: I'm willing to accept this for this PR, but I'd let to get a better understanding of what's going on with you sometime soon.

lifflander: I can write up a tiny example that shows the problem.

PhilMiller (approved):
Please address the wording, typo, and potentially the reconstruction ambiguity error check before merging.

The bigger things about the owning pointer helper and the registry definition can be dealt with later.

src/checkpoint/dispatch/dispatch_virtual.h
lifflander: Fixed

PR #107, opened 2020-05-28
Add a routine to serialize just the contents of a Kokkos::View, maintaining its identity

PhilMiller:
This can be used for an owning object to serialize the contents of View instances that have been reconstructed and aliased in multiple places.

### DIFF
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -424,6 +424,14 @@ void serializeExtentOnly(SerializerT& s, Kokkos::View<T**,Ts...>& v, std::string
   }
 }
 
+template< typename SerializerT, typename T, typename... Ts >
+void serializeContentsOnly(SerializerT& s, Kokkos::View<T, Ts...>& v) {
+  Kokkos::View<T> values = v;
+  s | values;
+  if (s.isUnpacking())
+    Kokkos::deep_copy(v, values);
+}
+
 #if KOKKOS_KERNELS_ENABLED
 template< typename Serializer, typename T, typename... Ts >
 inline void serialize( Serializer &s, Kokkos::StaticCrsGraph<T, Ts...> &graph ) {
--- a/tests/unit/test_kokkos_serialize_special.cc
+++ b/tests/unit/test_kokkos_serialize_special.cc
@@ -1,0 +1,83 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                     test_kokkos_serialize_special.cc
+//                           DARMA Toolkit v. 1.0.0
+//                 DARMA/checkpoint => Serialization Library
+//
+// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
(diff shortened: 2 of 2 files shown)


PhilMiller:
(Note: made in the Github UI, totally untested)

src/checkpoint/container/view_serialize.h
```diff
@@ -424,6 +424,14 @@ void serializeExtentOnly(SerializerT& s, Kokkos::View<T**,Ts...>& v, std::string
   }
 }
 
+template< typename SerializerT, typename T, typename... Ts >
+void serializeContentsOnly(SerializerT& s, Kokkos::View<T, Ts...>& v) {
```
uhetmaniuk, line 428: Great addition! 
With this addition, what should be the meaning of `s | v` ? 
I wonder whether the user will expect to get the contents of v serialized (at least, I was expecting it and Matt B. could not explain the 0 entries in the output view).

PhilMiller: Your issue is likely that two different objects both have serialization code that executes s|v on View instances that refer to a common underlying array. When deserialized, each of them gets back a fresh instance. Those serialization routines need to be coordinated, such that one object owns the view and takes responsibility for serializing it, and the others are coordinated to obtain the same reference.

The function I add here will have no impact on plain s|v serialization of a view - it only applies where explicitly called. 

In the use case of EMPIRE restoration, where the entire scaffolding of a job will be rebuilt from the input deck, it may be appropriate to pretty much uniformly call serializeContentsOnly or nothing at all on View members, trusting that the initialization has guaranteed that all objects have correctly aliased referenced to any Views of shared arrays.

uhetmaniuk: Sounds good. Thank you for clarifying.

uhetmaniuk (approved):

lifflander (approved):
Looks solid to me.

tests/unit/test_kokkos_serialize_special.cc
```diff
@@ -0,0 +1,111 @@
...
+  auto buffer = packer.extractPackedBuffer();
+  auto unpacker = Unpacker(buffer->getBuffer());
+
+  serializeContentsOnly(unpacker, original);
+
+  EXPECT_EQ(10, alias(0));
+  EXPECT_EQ(20, alias(1));
+}
+
+struct KokkosViewExtentTest : virtual testing::Test { };
```
lifflander, line 83: Why have two different structs for testing? You can group both under a single class with a different name so they used up grouped together in gtest

PhilMiller: My lack of familiarity with gtest. Monkey see, monkey do. Want me to 'fix' it?

lifflander: Sure, it will make it easier to pattern match out tests.

PhilMiller: Actually, in this case, I think it makes sense to keep these two separate. They test distinct subjects, and their successes and failures should be somewhat distinct.

PR #114, opened 2020-06-15
113 Fix virtual derived de-serialization bug

### LINKED ISSUE #113: Virtual de-serialize from derived type fails in object registry lookup

lifflander:
Fixed #113

### DIFF
--- a/tests/unit/test_virtual_serialize.cc
+++ b/tests/unit/test_virtual_serialize.cc
@@ -344,12 +344,24 @@ struct TestWrapper {
     auto vec_size = vec.size();
     s | vec_size;
     vec.resize(vec_size);
+    auto vec_derived_size = vec_derived.size();
+    s | vec_derived_size;
+    vec_derived.resize(vec_derived_size);
 
     for (auto&& elm : vec) {
       TestBase* base = elm.get();
       checkpoint::allocateConstructForPointer(s, base);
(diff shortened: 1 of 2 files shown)


tests/unit/test_virtual_serialize.cc
```diff
@@ -365,10 +377,15 @@ struct TestWrapper {
         vec.emplace_back(std::make_shared<TestDerived3>(TEST_CONSTRUCT{}));
       }
     }
+
+    for (std::size_t i = 0; i < vec_num_elms; i++) {
+      vec_derived.emplace_back(std::make_shared<TestDerived1>(TEST_CONSTRUCT{}));
```
PhilMiller: It would perhaps be good to use `TestDerived2` and `TestDerived3` in exercising the case we're concerned about, since 2 is actually intermediate between `TestBase` and 3

PhilMiller:
I agree that the test code as written does probe the failure as actually observed, though

tests/unit/test_virtual_serialize.cc
lifflander: Done

lifflander:
> I agree that the test code as written does probe the failure as actually observed, though

Ok, I switched it. I concur that using the middle class is a better test.

PhilMiller (approved):

PR #115, opened 2020-06-29
79 Serialize Kokkos::DynRankView<T>

### LINKED ISSUE #79: Write serializer for Kokkos::DynRankView

lifflander:
Fixes #79

### DIFF
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -322,7 +322,21 @@ inline void serialize_impl(SerializerT& s, Kokkos::DynRankView<T,Args...>& view)
       // Serialize the data directly out of the data buffer
       dispatch::serializeArray(s, view.data(), num_elms);
     } else {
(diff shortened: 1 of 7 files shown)


src/checkpoint/container/view_serialize.h
```diff
@@ -235,6 +245,88 @@ inline void serialize(
...
+        "Serializing Kokkos::DynRankView is only supported up to 7 dimensions"
+      );
+    }
+  }
+
+  // Serialize the total number of elements in the Kokkos::View
+  size_t num_elms = view.size();
+  s | num_elms;
+
+  // Serialize whether the view is contiguous or not. Is this required?
```
PhilMiller, line 304: Why wouldn't this be required, unless we removed the `if (is_contig)` below and just used `TraverseManual` unconditionally? Does `DynRankView` guarantee that it's contiguous? Does it hurt to just check and do the detailed thing, in case that changes?

PhilMiller (approved):
Ship it

PR #116, opened 2020-07-01
Explicitly default-initialize primitive types to quiet spurious warnings

PhilMiller:


src/checkpoint/container/view_traverse_manual.h
```diff
@@ -153,16 +153,9 @@ struct TraverseManual<SerializerT,ViewType,1> {
...
     for (SizeType i = 0; i < v.extent(0); i++) {
       if (s.isUnpacking()) {
-        #pragma GCC diagnostic push
-#if !defined(__has_warning)
-        #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#elif __has_warning("-Wmaybe-uninitialized")
-        #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
-        BaseType val;
+        BaseType val{};
```
lifflander, line 156: Could this cause problems with constructibility? What types are allowed in Kokkos?

PhilMiller:
It's exactly the same default construction we were already doing by
declaring an instance without initialization. It should only impact
primitive types.

lifflander (approved):

lifflander:
> It's exactly the same default construction we were already doing by declaring an instance without initialization. It should only impact primitive types.
> […](#)

Good point. I think we are fine. Merging

PR #124, opened 2020-09-02
#123 Implement memory footprint mode

### LINKED ISSUE #123: Implement memory footprint mode
Implement a new mode and possibly a new operator for computing the memory footprint that has overloads for no serializable but sizable elements like `std::function<>` or pointers (just count the point size and then follow it).

cz4rs:
fixes #123

### DIFF
--- a/src/checkpoint/checkpoint_api.impl.h
+++ b/src/checkpoint/checkpoint_api.impl.h
@@ -93,8 +93,10 @@ std::size_t getSize(T& target) {
 
 template <typename T>
 std::size_t getMemoryFootprint(T& target) {
-  // FIXME: use std::max to make sure that it's not smaller than sizeof(T)
-  return dispatch::Standard::footprint<T, Footprinter>(target);
+  return std::max(
+    dispatch::Standard::footprint<T, Footprinter>(target),
+    sizeof(target)
+  );
 }
 
 template <typename T>
--- a/src/checkpoint/checkpoint_api.impl.h
+++ b/src/checkpoint/checkpoint_api.impl.h
@@ -93,6 +93,7 @@ std::size_t getSize(T& target) {
 
 template <typename T>
 std::size_t getMemoryFootprint(T& target) {
+  // FIXME: use std::max to make sure that it's not smaller than sizeof(T)
   return dispatch::Standard::footprint<T, Footprinter>(target);
 }
 
--- a/src/checkpoint/checkpoint_api.impl.h
+++ b/src/checkpoint/checkpoint_api.impl.h
@@ -92,6 +92,11 @@ std::size_t getSize(T& target) {
 }
 
 template <typename T>
+std::size_t getMemoryFootprint(T& target) {
+  return dispatch::Standard::footprint<T, Footprinter>(target);
+}
+
+template <typename T>
 void serializeToFile(T& target, std::string const& file) {
   auto len = getSize<T>(target);
   dispatch::Standard::pack<T, PackerBuffer<buffer::IOBuffer>>(
--- a/src/checkpoint/container/array_serialize.h
+++ b/src/checkpoint/container/array_serialize.h
@@ -54,6 +54,7 @@ namespace checkpoint {
 
 template <typename Serializer, typename T, size_t N>
 void serialize(Serializer& s, std::array<T, N>& array) {
+  s.countBytes(array);
   dispatch::serializeArray(s, &array[0], array.size());
 }
 
--- a/src/checkpoint/container/array_serialize.h
+++ b/src/checkpoint/container/array_serialize.h
@@ -54,7 +54,6 @@ namespace checkpoint {
 
 template <typename Serializer, typename T, size_t N>
 void serialize(Serializer& s, std::array<T, N>& array) {
-  s.countBytes(array);
   dispatch::serializeArray(s, &array[0], array.size());
 }
 
--- a/src/checkpoint/container/chrono_serialize.h
+++ b/src/checkpoint/container/chrono_serialize.h
@@ -2,11 +2,11 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                 checkpoint.h
+//                             chrono_serialize.h
 //                           DARMA Toolkit v. 1.0.0
 //                 DARMA/checkpoint => Serialization Library
 //
-// Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
+// Copyright 2020 National Technology & Engineering Solutions of Sandia, LLC
 // (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
 // Government retains certain rights in this software.
 //
@@ -42,28 +42,21 @@
 //@HEADER
 */
 
-#if !defined INCLUDED_CHECKPOINT_CHECKPOINT_H
-#define INCLUDED_CHECKPOINT_CHECKPOINT_H
+#if !defined INCLUDED_CHECKPOINT_CONTAINER_CHRONO_SERIALIZE_H
+#define INCLUDED_CHECKPOINT_CONTAINER_CHRONO_SERIALIZE_H
 
-#include "checkpoint/serializers/serializers_headers.h"
-#include "checkpoint/dispatch/dispatch.h"
-#include "checkpoint/traits/serializable_traits.h"
+#include <chrono>
 
-#include "checkpoint/container/array_serialize.h"
-#include "checkpoint/container/chrono_serialize.h"
-#include "checkpoint/container/enum_serialize.h"
-#include "checkpoint/container/function_serialize.h"
-#include "checkpoint/container/list_serialize.h"
-#include "checkpoint/container/map_serialize.h"
-#include "checkpoint/container/queue_serialize.h"
-#include "checkpoint/container/raw_ptr_serialize.h"
-#include "checkpoint/container/string_serialize.h"
-#include "checkpoint/container/tuple_serialize.h"
-#include "checkpoint/container/vector_serialize.h"
-#include "checkpoint/container/unique_ptr_serialize.h"
-#include "checkpoint/container/view_serialize.h"
+#include "checkpoint/common.h"
 
-#include "checkpoint/checkpoint_api.h"
-#include "checkpoint/checkpoint_api.impl.h"
 
-#endif /*INCLUDED_CHECKPOINT_CHECKPOINT_H*/
+namespace checkpoint {
+
+template <typename Serializer, typename Rep, typename Period>
+void serialize(Serializer& s, std::chrono::duration<Rep, Period> d) {
+  s.countBytes(d);
+}
+
+} /* end namespace checkpoint */
+
+#endif /*INCLUDED_CHECKPOINT_CONTAINER_CHRONO_SERIALIZE_H*/
--- a/src/checkpoint/container/chrono_serialize.h
+++ b/src/checkpoint/container/chrono_serialize.h
@@ -45,15 +45,19 @@
 #if !defined INCLUDED_CHECKPOINT_CONTAINER_CHRONO_SERIALIZE_H
 #define INCLUDED_CHECKPOINT_CONTAINER_CHRONO_SERIALIZE_H
 
-#include <chrono>
-
 #include "checkpoint/common.h"
 
+#include <chrono>
 
 namespace checkpoint {
 
 template <typename Serializer, typename Rep, typename Period>
 void serialize(Serializer& s, std::chrono::duration<Rep, Period> d) {
+  checkpointAssert(
+    s.isFootprinting(),
+    "Only memory footprinting is supported for this type"
+  );
+
   s.countBytes(d);
 }
 
--- a/src/checkpoint/container/chrono_serialize.h
+++ b/src/checkpoint/container/chrono_serialize.h
@@ -52,13 +52,10 @@
 namespace checkpoint {
 
 template <typename Serializer, typename Rep, typename Period>
-void serialize(Serializer& s, std::chrono::duration<Rep, Period> d) {
-  checkpointAssert(
-    s.isFootprinting(),
-    "Only memory footprinting is supported for this type"
-  );
-
-  s.countBytes(d);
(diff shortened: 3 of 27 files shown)


cz4rs:
note:
pushing some code here, but it's definitely still work-in-progress (requires improved naming (!) and docs, reduced code duplication, independence from `serialize` / separate operator, implementation for other containers etc.)

cz4rs:
@lifflander I have removed the custom dispatcher and used existing `serialize` code when possible, can you have a look?

- standard containers, fundamental types, enums seem to be working fine, code for `std::tuple`, `std::function` and other necessary types will land tomorrow
- the tests are failing currently because of how raw pointers are handled (currently just counting the pointer size and following it, but that doesn't work for C-style arrays and C strings) - any idea how to proceed with this?
- I have updated https://github.com/DARMA-tasking/vt/pull/1013 to have an example on the `vt` side, if that looks ok I will start working on components one by one to see if any issues come out
- how much testing is required for when `detector` is not available? (currently the CI only tests with `detector` present)

src/checkpoint/container/array_serialize.h
```diff
@@ -54,6 +54,7 @@ namespace checkpoint {
 
 template <typename Serializer, typename T, size_t N>
 void serialize(Serializer& s, std::array<T, N>& array) {
+  s.countBytes(array);
   dispatch::serializeArray(s, &array[0], array.size());
 }
```
PhilMiller: I think this is double-counting, since the implementation is required to contain the `T[N]` as a member, not point to it. So, I think the `countBytes` can just be omitted

src/checkpoint/container/chrono_serialize.h
```diff
@@ -0,0 +1,62 @@
...
+#include <chrono>
+
+#include "checkpoint/common.h"
+
+
+namespace checkpoint {
+
+template <typename Serializer, typename Rep, typename Period>
+void serialize(Serializer& s, std::chrono::duration<Rep, Period> d) {
+  s.countBytes(d);
```
PhilMiller: Since this can only compute a footprint, it should probably assert that the serializer is in that mode, rather than silently failing

src/checkpoint/container/queue_serialize.h
```diff
@@ -0,0 +1,79 @@
...
+namespace checkpoint {
+
+template <typename Serializer, typename T>
+void serialize(Serializer& s, std::queue<T> q) {
+  s.countBytes(q);
+
+  while (not q.empty())
+  {
+    s | q.front();
+    q.pop();
```
PhilMiller: These implementations are destructive, which is quite undesirable.

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * For footprinting mode, count the pointer size and follow it (note that it
+ * doesn't work correctly for C-style arrays!).
+ *
+ * \param[in] serializer serializer to use
+ * \param[in] ptr pointer to serialize
+ */
+template <typename Serializer, typename T>
+void serialize(Serializer& s, T* ptr) {
+  s.countBytes(ptr);
```
PhilMiller: Again, since this will only work for footprinting, it should assert `s.isFootprinting()`

src/checkpoint/container/shared_ptr_serialize.h
```diff
@@ -0,0 +1,63 @@
...
+
+#if !defined INCLUDED_CHECKPOINT_CONTAINER_SHARED_PTR_SERIALIZE_H
+#define INCLUDED_CHECKPOINT_CONTAINER_SHARED_PTR_SERIALIZE_H
+
+#include "checkpoint/common.h"
+
+namespace checkpoint {
+
+template <typename Serializer, typename T>
+void serialize(Serializer& s, std::shared_ptr<T>& ptr) {
```
PhilMiller: Ditto about assertion

src/checkpoint/container/string_serialize.h
```diff
@@ -61,8 +61,13 @@ void serializeStringMeta(Serializer& s, std::string& str) {
 
 template <typename Serializer>
 void serialize(Serializer& s, std::string& str) {
-  serializeStringMeta(s, str);
-  dispatch::serializeArray(s, str.c_str(), str.size());
+  if (s.isFootprinting()) {
+    s.countBytes(str);
+    dispatch::serializeArray(s, str.c_str(), str.capacity());
```
PhilMiller, line 75: This too potentially runs into the small-array optimization, but I think it's *not* a problem here, because `std::string` has a *fixed* small array size, so the overcount is tightly limited

PhilMiller:
Overall, I like this.

We'll want to figure out some systematic way to identify whether a class needs special footprint handling, and whether we've considered the need for special handling in each class for which we have a serializer.

Since we're handling container classes of various sorts in checkpoint, the serializers for `Kokkos` types need to be considered, too. I'm a little concerned that we don't correctly handle the special case of `Kokkos::View<Kokkos::View<T*>>` - the nested views will take up space even if empty, and need to be counted.

src/checkpoint/container/queue_serialize.h
cz4rs: I couldn't find a better (concise) way to iterate over queue's elements, so I went with passing by value - `pop` is called on the copy. This is not too elegant and potentially slow, but the original object remains untouched.

src/checkpoint/checkpoint_api.impl.h
```diff
@@ -91,6 +91,12 @@ std::size_t getSize(T& target) {
   return dispatch::Standard::size<T, Sizer>(target);
 }
 
+template <typename T>
+std::size_t getMemoryFootprint(T& target) {
+  // FIXME: use std::max to make sure that it's not smaller than sizeof(T)
```
cz4rs: at this stage (of implementation) `getMemoryFootprint` sometimes returns smaller value than a simple `sizeof`
(this can be valid calculation because of padding / layout)

note to self: change to `std::max(footprint, sizeof(T))` before merging

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -0,0 +1,89 @@
...
+ * \brief Serialize raw pointer \c ptr
+ *
+ * For footprinting mode, count the pointer size and follow it (note that it
+ * doesn't work correctly for C-style arrays!).
+ *
+ * \param[in] serializer serializer to use
+ * \param[in] ptr pointer to serialize
+ */
+template <typename Serializer, typename T>
+void serialize(Serializer& s, T* ptr) {
```
lifflander: I think we need to bifurcate the dispatch for non-serializable, but footprintable types. Or cases where we want footprinting to be handled quite differently than regular serialization  We could do that statically, with an enable_if on the `SerializerT`. See comment at the bottom.

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -0,0 +1,89 @@
...
+
+  s.countBytes(ptr);
+  if (ptr != nullptr) {
+    s | *ptr;
+  }
+}
+
+template <typename Serializer>
+void serialize(Serializer& s, FILE* ptr) {
+  checkpointAssert(
```
lifflander: Same here.

src/checkpoint/container/chrono_serialize.h
lifflander: For this one, can we add a real serializer also here? I don't think chrono should be that difficult to serialize

src/checkpoint/container/queue_serialize.h
lifflander: It's quite unfortunate that this container (queue) doesn't have access to the underlying data structure.

lifflander:
@cz4rs @PhilMiller I'd like to include enable_if's on footprinting-only serializers, so that they can't accidentally be used by other serializers. This also makes it clear why the code doesn't compile if some footprint specific member (like `s.countBytes`) is called and then fails to work with Packer/Unpacker/Sizer.

```C++
template <
  typename SerializerT,
  typename T,
  typename = std::enable_if_t<std::is_same<SerializerT, checkpoint::Footprinter>::value>
>
void serialize(SerializerT& s, std::queue<T> queue) {

}
```

If we want to make this code more general (assuming different footprinters might be used), we might want to check for a type trait instead of the exact type `Footprinter`.

src/checkpoint/container/raw_ptr_serialize.h
cz4rs: enforcing Footprinting mode with `std::enable_if_t`

cz4rs:
> I'd like to include enable_if's on footprinting-only serializers, so that they can't accidentally be used by other serializers. This also makes it clear why the code doesn't compile if some footprint specific member (like `s.countBytes`) is called and then fails to work with Packer/Unpacker/Sizer.

I have implemented `enable_if`'s in a helper function called by `serialize` and using anything else than `getMemoryFootprint` fails during compilation as expected. Putting `enable_if` directly in `serialize` was causing problems with proper detection in `dispatch_serializer_nonbyte.h`.

src/checkpoint/container/shared_ptr_serialize.h
cz4rs: `std_enable_if_t` added

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include "checkpoint/common.h"
+
+namespace checkpoint {
+
+/**
+ * \brief Serialize raw pointer \c ptr
+ *
+ * Only footprinting mode is supported at the moment. Counts the pointer size
+ * and follows it (note that it doesn't work correctly for C-style arrays!).
```
cz4rs, line 56: handling arrays correctly could be possible using `std::is_array` or
```cpp
template<typename SerializerT, typename T, size_t N> 
void serialize(T (&t)[N])
```

I poked around it a little bit, but currently arrays decay into pointers in the middle of call stack and I have left the warning here

src/checkpoint/container/array_serialize.h
cz4rs: removed the redundant `countBytes` call

src/checkpoint/container/chrono_serialize.h
cz4rs: serializer added

src/checkpoint/container/queue_serialize.h
```diff
@@ -0,0 +1,104 @@
...
+  typename SerializerT,
+  typename T,
+  typename = std::enable_if_t<
+    std::is_same<
+      SerializerT,
+      checkpoint::Footprinter
+    >::value
+  >
+>
+void serializeQueue(SerializerT& s, std::queue<T> q) {
```
PhilMiller: Per discussion on today's call, we prefer to just count `q.size() * sizeof(T)`, and sacrifice some accuracy to avoid the potential pitfall of copying and iterating over every element.

PhilMiller: That would shift us back to taking the queue arguments by `const&`

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -0,0 +1,116 @@
...
+    std::is_same<
+      SerializerT,
+      checkpoint::Footprinter
+    >::value
+  >
+>
+void serializeFilePtr(SerializerT& s, FILE* ptr) {
+  s.countBytes(ptr);
+  if (ptr != nullptr) {
+    s.countBytes(*ptr);
```
PhilMiller: On some platforms, I believe `struct FILE;` is left as an incomplete definition, to avoid exposing the C library's implementation. I think this would fail to compile in those settings, since it's ultimately going to attempt `sizeof()` on an incomplete type.

src/checkpoint/container/string_serialize.h
```diff
@@ -59,10 +58,25 @@ void serializeStringMeta(Serializer& s, std::string& str) {
   str.resize(str_size);
 }
 
+/**
+ * \brief Serialize string \c str
+ *
+ * Resizes a string to it's actual size and serializes it.
```
PhilMiller: Typo - should be "its"

src/checkpoint/container/raw_ptr_serialize.h
cz4rs: changed to size the pointer only

src/checkpoint/container/string_serialize.h
cz4rs: fixed

src/checkpoint/container/queue_serialize.h
cz4rs: done, the inaccuracy will be bigger when the underlying container is `std::vector` (default for priority queue), but there's no access to capacity in that case

cz4rs:
@PhilMiller @lifflander if this looks ok for you, I would like to get it merged to have better CI feedback in https://github.com/DARMA-tasking/vt/pull/1013

(I would be adding support for kokkos types and ironing out any discovered bugs in a subsequent PR)

PhilMiller (approved):

src/checkpoint/container/queue_serialize.h
PhilMiller: I realize that for the specific cases in the runtime of these containers, we can handle footprinting specially, by keeping a running measure of the footprint of the contents, to add to the footprint explicitly.

Which means, the footprint interface needs to be able to accept "add arbitrary byte count"

cz4rs:
@lifflander @PhilMiller (thanks for the review!)
regarding the merge permissions / protected branches: please take a look at the screenshot below, I cannot merge this pull request myself
- the build passed fine
- review is approved "by reviewer with write access"
- the "not authorized" comment links to https://docs.github.com/en/github/administering-a-repository/about-protected-branches

I don't have a fresh example from `vt` repository, but it looked pretty similar there (we discussed this some time ago on DARMA technical meeting)

image

lifflander:
> @lifflander @PhilMiller (thanks for the review!)
> regarding the merge permissions / protected branches: please take a look at the screenshot below, I cannot merge this pull request myself

I believe that I have fixed this now for this repository.

PR #126, opened 2020-09-22
123 add support for more types in memory footprinting mode

### LINKED ISSUE #123: Implement memory footprint mode

cz4rs:
adding memory footprinting support for `vt` components (https://github.com/DARMA-tasking/vt/issues/1009)

fixes #123


cz4rs:
move the change into a new issue

PR #128, opened 2020-09-22
127 Add support for more types in memory footprinting mode

### LINKED ISSUE #127: Add support for more types in memory footprinting mode
adding memory footprinting for vt components (DARMA-tasking/vt#1009) requires more types to be supported:

- [x] `std::stack`
- [x] `void*`
- [x] `std::map` with `std::unique_ptr` mapped type 
- [x] `std::_List_iterator<T>`
- [x] ~~non-intrusive serializer for `omp_lock_t` (and any other OpenMP types used)~~
note: just count the size without diving into implementation
- [x] ~~handle pure abstract interface (case of `vt::vrt::collection::balance::LoadModel`)~~
note: this should be handled on the `vt` side using virtual serialization
- [x] ~~handle ambiguous case, when a type both `hasNoninSerialize`, `isEnum` (`enum class vt::vrt::collection::balance::LBType`)~~
note: this should be handled on `vt` side, remove custom `serialize` for enum class

note: this is a followup on #123

cz4rs:
fixes #127

### DIFF
--- a/src/checkpoint/container/raw_ptr_serialize.h
+++ b/src/checkpoint/container/raw_ptr_serialize.h
@@ -65,13 +65,13 @@ void serialize(SerializerT& s, T* ptr) {
 
 template <
   typename SerializerT,
-  typename T,
-  typename = std::enable_if_t<
-    std::is_same<
-      SerializerT,
-      checkpoint::Footprinter
-    >::value
-  >
+  typename T//,
+  // typename = std::enable_if_t<
+  //   std::is_same<
+  //     SerializerT,
+  //     checkpoint::Footprinter
+  //   >::value
+  // >
 >
 void serializeRawPtr(SerializerT& s, T* ptr) {
   s.countBytes(ptr);
--- a/src/checkpoint/container/raw_ptr_serialize.h
+++ b/src/checkpoint/container/raw_ptr_serialize.h
@@ -58,20 +58,23 @@ namespace checkpoint {
  * \param[in] serializer serializer to use
  * \param[in] ptr pointer to serialize
  */
-template <typename SerializerT, typename T>
+template <
+  typename SerializerT,
+  typename T,
+  typename = std::enable_if_t<
+    std::is_same<
+      SerializerT,
+      checkpoint::Footprinter
+    >::value
+  >
+>
 void serialize(SerializerT& s, T* ptr) {
   serializeRawPtr(s, ptr);
 }
 
 template <
   typename SerializerT,
-  typename T//,
-  // typename = std::enable_if_t<
-  //   std::is_same<
-  //     SerializerT,
-  //     checkpoint::Footprinter
-  //   >::value
-  // >
+  typename T
 >
 void serializeRawPtr(SerializerT& s, T* ptr) {
   s.countBytes(ptr);
@@ -84,13 +87,7 @@ void serializeRawPtr(SerializerT& s, T* ptr) {
  * Note: do not follow void pointer.
  */
 template <
-  typename SerializerT,
-  typename = std::enable_if_t<
-    std::is_same<
-      SerializerT,
-      checkpoint::Footprinter
-    >::value
-  >
+  typename SerializerT
 >
 void serializeRawPtr(SerializerT& s, void* ptr) {
   s.countBytes(ptr);
@@ -101,13 +98,7 @@ void serializeRawPtr(SerializerT& s, void* ptr) {
  * platforms.
  */
 template <
-  typename SerializerT,
-  typename = std::enable_if_t<
-    std::is_same<
-      SerializerT,
-      checkpoint::Footprinter
-    >::value
-  >
+  typename SerializerT
 >
 void serializeRawPtr(SerializerT& s, FILE* ptr) {
   s.countBytes(ptr);
--- a/src/checkpoint/container/raw_ptr_serialize.h
+++ b/src/checkpoint/container/raw_ptr_serialize.h
@@ -81,19 +81,8 @@ void serializeRawPtr(SerializerT& s, T* ptr) {
 }
 
 /**
- * \brief Serialize FILE pointer \c ptr
- *
- * Only footprinting mode is supported at the moment. Counts the pointer size
- * without following it.
- *
- * \param[in] serializer serializer to use
- * \param[in] ptr pointer to serialize
+ * Note: do not follow void pointer.
  */
-template <typename SerializerT>
-void serialize(SerializerT& s, FILE* ptr) {
-  serializeFilePtr(s, ptr);
-}
-
 template <
   typename SerializerT,
   typename = std::enable_if_t<
@@ -103,7 +92,24 @@ template <
     >::value
   >
 >
-void serializeFilePtr(SerializerT& s, FILE* ptr) {
+void serializeRawPtr(SerializerT& s, void* ptr) {
+  s.countBytes(ptr);
+}
+
+/**
+ * Note: do not follow FILE pointer as it might be an incomplete type on some
+ * platforms.
+ */
+template <
+  typename SerializerT,
+  typename = std::enable_if_t<
+    std::is_same<
+      SerializerT,
+      checkpoint::Footprinter
+    >::value
+  >
+>
+void serializeRawPtr(SerializerT& s, FILE* ptr) {
   s.countBytes(ptr);
 }
 
--- a/src/checkpoint/container/vector_serialize.h
+++ b/src/checkpoint/container/vector_serialize.h
@@ -66,18 +66,16 @@ template <typename SerializerT, typename T, typename VectorAllocator>
 typename std::enable_if_t<
   std::is_same<SerializerT, checkpoint::Footprinter>::value,
   void
-> serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) { }
+> serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) {
+  s.countBytes(vec);
+}
 
 
 template <typename Serializer, typename T, typename VectorAllocator>
 void serialize(Serializer& s, std::vector<T, VectorAllocator>& vec) {
-  if (s.isFootprinting()) {
-    s.countBytes(vec);
-    dispatch::serializeArray(s, &vec[0], vec.capacity());
-  } else {
-    serializeVectorMeta(s, vec);
-    dispatch::serializeArray(s, &vec[0], vec.size());
-  }
+  serializeVectorMeta(s, vec);
+  std::size_t num_elements = s.isFootprinting() ? vec.capacity() : vec.size();
+  dispatch::serializeArray(s, &vec[0], num_elements);
 }
 
 template <typename Serializer, typename VectorAllocator>
--- a/src/checkpoint/container/vector_serialize.h
+++ b/src/checkpoint/container/vector_serialize.h
@@ -52,12 +52,22 @@
 
 namespace checkpoint {
 
-template <typename Serializer, typename T, typename VectorAllocator>
-void serializeVectorMeta(Serializer& s, std::vector<T, VectorAllocator>& vec) {
+template <typename SerializerT, typename T, typename VectorAllocator>
+typename std::enable_if_t<
+  not std::is_same<SerializerT, checkpoint::Footprinter>::value,
+  void
+> serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) {
   SerialSizeType vec_size = vec.size();
   s | vec_size;
   vec.resize(vec_size);
 }
+
+template <typename SerializerT, typename T, typename VectorAllocator>
+typename std::enable_if_t<
+  std::is_same<SerializerT, checkpoint::Footprinter>::value,
+  void
+> serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) { }
+
 
 template <typename Serializer, typename T, typename VectorAllocator>
 void serialize(Serializer& s, std::vector<T, VectorAllocator>& vec) {
--- a/src/checkpoint/serializers/footprinter.h
+++ b/src/checkpoint/serializers/footprinter.h
@@ -65,6 +65,10 @@ struct Footprinter : Serializer {
     num_bytes_ += sizeof(t);
   }
 
+  void addBytes(std::size_t s) {
+    num_bytes_ += s;
+  }
+
 private:
   SerialSizeType num_bytes_ = 0;
 };
--- a/tests/unit/test_footprinter.cc
+++ b/tests/unit/test_footprinter.cc
@@ -47,6 +47,7 @@
 #include "test_harness.h"
 
 #include <atomic>
+#include <thread>
 #include <checkpoint/checkpoint.h>
 
 namespace checkpoint { namespace tests { namespace unit {
@@ -396,7 +397,17 @@ TEST_F(TestFootprinter, test_string) {
   std::string s = "123456789";
   EXPECT_EQ(
     checkpoint::getMemoryFootprint(s),
-    sizeof(s) + s.capacity() * sizeof(s[0]));
+    sizeof(s) + s.capacity() * sizeof(s[0])
+  );
+}
+
+TEST_F(TestFootprinter, test_thread) {
+  std::thread t(fn);
+  EXPECT_EQ(
+    checkpoint::getMemoryFootprint(t),
+    sizeof(t)
+  );
+  t.detach();
 }
 
 // naive approach, just sum memory usage of all elements
--- a/tests/unit/test_footprinter.cc
+++ b/tests/unit/test_footprinter.cc
@@ -328,6 +328,20 @@ TEST_F(TestFootprinter, test_set) {
   }
 }
 
+TEST_F(TestFootprinter, test_stack) {
+  {
+    std::stack<int> stack;
+    stack.push(1);
+    stack.push(2);
+    stack.push(3);
+    EXPECT_EQ(
+      checkpoint::getMemoryFootprint(stack),
+      sizeof(stack) + stack.size() * sizeof(stack.top())
+    );
+    EXPECT_EQ(stack.size(), std::size_t{3});
+  }
+}
+
 // does not account for small string optimisation
 TEST_F(TestFootprinter, test_string) {
   std::string s = "123456789";
--- a/tests/unit/test_footprinter.cc
+++ b/tests/unit/test_footprinter.cc
@@ -476,7 +476,15 @@ struct TestDerived2 : TestBase {
 
   checkpoint_virtual_serialize_derived(TestDerived2, TestBase)
 
-  template <typename SerializerT>
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<
+        SerializerT,
+        checkpoint::Footprinter
+      >::value
+    >
+  >
   void serialize(SerializerT& s) {
     s | raw_pointer;
     s | shared_pointer;
--- a/tests/unit/test_footprinter.cc
+++ b/tests/unit/test_footprinter.cc
@@ -473,6 +473,8 @@ private:
 };
 
 struct TestDerived2 : TestBase {
+  explicit TestDerived2(int i) {}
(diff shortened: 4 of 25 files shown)


lifflander (approved):
This is looking good to me. We definitely should make sure that all the serializers work with `std::unique_ptr` as much as possible. That might have been overlooked since support for normal serialization of `std::unique_ptr` was implemented fairly recently.

lifflander:
I'm not going to merge this because I'm not sure if you are done yet.

cz4rs:
@lifflander 
> I'm not going to merge this because I'm not sure if you are done yet.

there is a bit of work to support remaining required types, but the bigger challenge that I'm facing right now is with virtual serialization
- I have started to implement virtual serialization for `LoadModel` on vt side, based on examples and comments in `dispatch_virtual.h`
- I've had to take a step back with 8157461  - if shared_ptr's `serialize` is available only to `Footprinter`, I get:
```
[build] ../lib/checkpoint/src/checkpoint/container/shared_ptr_serialize.h: In instantiation of ‘void checkpoint::serialize(SerializerT&, std::shared_ptr<_Tp>&) [with SerializerT = checkpoint::PackerBuffer<checkpoint::buffer::ManagedBuffer>; T = vt::vrt::collection::balance::LoadModel]’:
[build] ../lib/checkpoint/src/checkpoint/dispatch/dispatch_serializer_nonbyte.h:86:14:   required from ‘static void checkpoint::dispatch::BasicDispatcher<SerializerT, T>::serializeNonIntrusive(SerializerT&, T&) [with SerializerT = checkpoint::PackerBuffer<checkpoint::buffer::ManagedBuffer>; T = std::shared_ptr<vt::vrt::collection::balance::LoadModel>]’
(...)
[build] ../lib/checkpoint/src/checkpoint/dispatch/vrt/serialize_instantiator.h:91:80:   required from ‘void checkpoint::instantiateObjSerializer() [with ObjT = vt::vrt::collection::balance::PerCollection; SerializerTs = {checkpoint::Footprinter, checkpoint::PackerBuffer<checkpoint::buffer::ManagedBuffer>, checkpoint::PackerBuffer<checkpoint::buffer::UserBuffer>, checkpoint::PackerBuffer<checkpoint::buffer::IOBuffer>, checkpoint::UnpackerBuffer<checkpoint::buffer::UserBuffer>, checkpoint::UnpackerBuffer<checkpoint::buffer::IOBuffer>, checkpoint::Sizer}]’
[build] ../src/vt/vrt/collection/balance/model/per_collection.h:69:3:   required from here
[build] ../lib/checkpoint/src/checkpoint/container/shared_ptr_serialize.h:54:21: error: no matching function for call to ‘serializeSharedPtr(checkpoint::PackerBuffer<checkpoint::buffer::ManagedBuffer>&, std::shared_ptr<vt::vrt::collection::balance::LoadModel>&)’
[build]    54 |   serializeSharedPtr(s, ptr);
```
(as you can see, `checkpoint::PackerBuffer` is used along the way)
- the other issue that I haven't figured out yet on `vt` side (https://github.com/DARMA-tasking/vt/pull/1013/commits/98360e8aa6618b6fdc3ed46845c532b0f861a697)
```
[build] ../lib/checkpoint/src/checkpoint/dispatch/dispatch_serializer_nonbyte.h: In instantiation of ‘void checkpoint::dispatch::SerializerDispatchNonByte<SerializerT, T, Dispatcher>::apply(SerializerT&, T*, checkpoint::SerialSizeType, checkpoint::dispatch::SerializerDispatchNonByte<SerializerT, T, Dispatcher>::hasNotVirtualSerialize<U>*) [with U = vt::vrt::collection::balance::LoadModel; SerializerT = checkpoint::Footprinter; T = vt::vrt::collection::balance::LoadModel; Dispatcher = checkpoint::dispatch::BasicDispatcher<checkpoint::Footprinter, vt::vrt::collection::balance::LoadModel>; checkpoint::SerialSizeType = long unsigned int; checkpoint::dispatch::SerializerDispatchNonByte<SerializerT, T, Dispatcher>::hasNotVirtualSerialize<U> = vt::vrt::collection::balance::LoadModel]’:
(...)
[build] ../src/vt/vrt/collection/balance/lb_invoke/lb_manager.h:265:7:   required from ‘void vt::vrt::collection::balance::LBManager::serialize(Serializer&) [with Serializer = checkpoint::Footprinter]’
(...)
[build] ../lib/checkpoint/src/checkpoint/checkpoint_api.impl.h:97:50:   required from ‘std::size_t checkpoint::getMemoryFootprint(T&) [with T = vt::vrt::collection::balance::LBManager; std::size_t = long unsigned int]’
[build] ../tests/unit/runtime/test_memory_footprint.cc:83:48:   required from ‘void vt::tests::unit::printMemoryFootprint(T*) [with T = vt::vrt::collection::balance::LBManager]’
[build] ../tests/unit/runtime/test_memory_footprint.cc:96:38:   required from here
[build] ../lib/checkpoint/src/checkpoint/dispatch/dispatch_serializer_nonbyte.h:218:5: error: no matching function for call to ‘checkpoint::dispatch::SerializerDispatchNonByte<checkpoint::Footprinter, vt::vrt::collection::balance::LoadModel, checkpoint::dispatch::BasicDispatcher<checkpoint::Footprinter, vt::vrt::collection::balance::LoadModel> >::applyStatic(checkpoint::Footprinter&, vt::vrt::collection::balance::LoadModel*&, checkpoint::SerialSizeType&)’
[build]   218 |     applyStatic(s, val, num);
[build]       |     ^~~~~~~~~~~
[build] ../lib/checkpoint/src/checkpoint/dispatch/dispatch_serializer_nonbyte.h:163:8: note: candidate: ‘template<class U> void checkpoint::dispatch::SerializerDispatchNonByte<SerializerT, T, Dispatcher>::applyStatic(SerializerT&, T*, checkpoint::SerialSizeType, checkpoint::dispatch::SerializerDispatchNonByte<SerializerT, T, Dispatcher>::hasInSerialize<U>*) [with U = U; SerializerT = checkpoint::Footprinter; T = vt::vrt::collection::balance::LoadModel; Dispatcher = checkpoint::dispatch::BasicDispatcher<checkpoint::Footprinter, vt::vrt::collection::balance::LoadModel>]’
[build]   163 |   void applyStatic(
[build]       |        ^~~~~~~~~~~
(...)
[build] ../lib/checkpoint/src/checkpoint/dispatch/dispatch_serializer_nonbyte.h:114:11: error: no type named ‘type’ in ‘struct std::enable_if<false, vt::vrt::collection::balance::LoadModel>’
[build]   114 |     using hasInSerialize =
```
the intrusive `serialize` method in `LoadModel` is not detected for some reason - do you have any idea why? as far as I can see the implementation seems correct

cz4rs:
if you prefer clang output, build log is available (for example) here: https://github.com/DARMA-tasking/vt/pull/1013/checks?check_run_id=1162292999
```
/build/checkpoint/install/include/checkpoint/dispatch/dispatch_serializer_nonbyte.h:163:8: note: candidate template ignored: requirement 'SerializableTraits<vt::vrt::collection::balance::LoadModel>::has_serialize_instrusive' was not satisfied [with U = vt::vrt::collection::balance::LoadModel]
  void applyStatic(
       ^
```

lifflander:
> * I've had to take a step back with [e5fe0f2](https://github.com/DARMA-tasking/checkpoint/commit/e5fe0f2e4b11694dff092a9b0b3c56c30dea9881) - if shared_ptr's `serialize` is available only to `Footprinter`, I get:
> (as you can see, `checkpoint::PackerBuffer` is used along the way)

For virtual serialize to work properly, you need to define your serializer as part of a registration macro. Add to this list:

https://github.com/DARMA-tasking/checkpoint/blob/7c8115e7d63eef9cc8dcaaf8f2c9c5fd31089cae/src/checkpoint/serializers/serializers_headers.h#L55-L61

However, I think it's a bug that it doesn't tell you it wasn't registered properly. It should check that when you use virtual serialize. But, once you add it to that list, it should be dispatched properly.

We should also document this. All traversers/serializers should be added to this list (if in checkpoint), or should call another API if they are user-written traversals for virtual serialize to work properly. The reason it has to be in this list is because we have to instantiate the `SerializerT` when virtual dispatch happens.

> * the other issue that I haven't figured out yet on `vt` side ([DARMA-tasking/vt@98360e8](https://github.com/DARMA-tasking/vt/commit/98360e8aa6618b6fdc3ed46845c532b0f861a697))
> the intrusive `serialize` method in `LoadModel` is not detected for some reason - do you have any idea why? as far as I can see the implementation seems correct

I'm looking into this now. Not sure off hand.

** Also there is a page that discuss virtual serializers in some detail **
https://github.com/DARMA-tasking/checkpoint/blob/develop/src/checkpoint/dispatch/dispatch_virtual.h

I think that using the macros for virtual serialize might be cleaner than the shim classes in the `LoadModel` case so we don't take over the inheritance hierarchy.

lifflander:
(I'm making this a draft so it doesn't get merged accidentally before it is complete)

cz4rs:
> For virtual serialize to work properly, you need to define your serializer as part of a registration macro. Add to this list:

Yep, `Footprinter` serializer was missing on that list, but the logs were produced after I have added it with efe836cd842645c5db95cfd0d30a29708b7c2416


> (I'm making this a draft so it doesn't get merged accidentally before it is complete)

Oops, I forgot to mark it as a draft myself - sorry about that!

cz4rs:
@PhilMiller (following from the other issue)
> In the scheduler or active messenger, newly-enqueued messages should have their size added to some gauge variable to track their total footprint, and be subtracted when they're processed. That variable's value can then be added to the scheduler component's footprint when that's being calculated. If there's no method on the footprinter to add an arbitrary value to the computed footprint, that should get implemented.

can it be as simple as d2a8c37?

lifflander:
> > For virtual serialize to work properly, you need to define your serializer as part of a registration macro. Add to this list:
> 
> Yep, `Footprinter` serializer was missing on that list, but the logs were produced after I have added it with [efe836c](https://github.com/DARMA-tasking/checkpoint/commit/efe836cd842645c5db95cfd0d30a29708b7c2416)
> 
> > (I'm making this a draft so it doesn't get merged accidentally before it is complete)
> 
> Oops, I forgot to mark it as a draft myself - sorry about that!

Sorry, I missed that it was part of the list. Probably had an old version of the branch.

lifflander:
> @PhilMiller (following from the other issue)
> 
> > In the scheduler or active messenger, newly-enqueued messages should have their size added to some gauge variable to track their total footprint, and be subtracted when they're processed. That variable's value can then be added to the scheduler component's footprint when that's being calculated. If there's no method on the footprinter to add an arbitrary value to the computed footprint, that should get implemented.
> 
> can it be as simple as [d2a8c37](https://github.com/DARMA-tasking/checkpoint/commit/d2a8c37a0ec20941fcb36443e1d4ca3a1b4d16f8)?

That looks reasonable to me if we only want to add it to the total sum, but we might want something that can be called in `serialize` recursively on the footprinter `SerializerT`. 

I'm not sure how we want to support that when the `SerializerT` is not in footprinting mode and we aren't using an enable_if.

lifflander:
> > For virtual serialize to work properly, you need to define your serializer as part of a registration macro. Add to this list:
> 
> Yep, `Footprinter` serializer was missing on that list, but the logs were produced after I have added it with [efe836c](https://github.com/DARMA-tasking/checkpoint/commit/efe836cd842645c5db95cfd0d30a29708b7c2416)
> 
> > (I'm making this a draft so it doesn't get merged accidentally before it is complete)
> 
> Oops, I forgot to mark it as a draft myself - sorry about that!

I've taken a look and don't see anything obviously wrong. Clang seems to get the Footprinter right at least (not `PackerBuffer`) if I'm reading it correctly.

I think the dispatch traits might not be triggering for some reason. Do you have an overload for shared_ptr that calls properly for virtual serialize when footprinting?

lifflander:
Checkpoint by default does not provide a serialize for shared_ptr on purpose. This is because it often has shared ownership so naively serializing it can end up with multiple copies and broken references during de-serialization.

cz4rs:
> Checkpoint by default does not provide a serialize for shared_ptr on purpose. This is because it often has shared ownership so naively serializing it can end up with multiple copies and broken references during de-serialization.

so I understand that keeping the `enable_if` is desirable (to make it work only with Footprinter)

lifflander:
> > Checkpoint by default does not provide a serialize for shared_ptr on purpose. This is because it often has shared ownership so naively serializing it can end up with multiple copies and broken references during de-serialization.
> 
> so I understand that keeping the `enable_if` is desirable (to make it work only with Footprinter)

Yes, exactly.

cz4rs:
8157461 adds a testcase displaying the problems with virtual serialization and `shared_ptr`, to isolate it / make it easier to analyze (it fails, because of `enable_if` working as intended)

I am currently analyzing `serialize_instantiator.h` (and the rest of the `vrt`) to see why `PackerBuffer` is picked from `SerializerTs`, where we have all the serializers present (including `Footprinter`)
```
../src/checkpoint/dispatch/vrt/serialize_instantiator.h:91:80:   required from ‘void checkpoint::instantiateObjSerializer() [with ObjT = checkpoint::tests::unit::test_3::TestDerived2; SerializerTs = {checkpoint::Footprinter, checkpoint::PackerBuffer<checkpoint::buffer::ManagedBuffer>, checkpoint::PackerBuffer<checkpoint::buffer::UserBuffer>, checkpoint::PackerBuffer<checkpoint::buffer::IOBuffer>, checkpoint::UnpackerBuffer<checkpoint::buffer::UserBuffer>, checkpoint::UnpackerBuffer<checkpoint::buffer::IOBuffer>, checkpoint::Sizer}]’
../tests/unit/test_virtual_serialize.cc:503:3:   required from here
../src/checkpoint/container/shared_ptr_serialize.h:54:21: error: no matching function for call to ‘serializeSharedPtr(checkpoint::PackerBuffer<checkpoint::buffer::ManagedBuffer>&, std::shared_ptr<int>&)’
```

edit: using `checkpoint::allocateConstructForPointer` seems like the way to go, but I haven't got it to work yet (although in this case there is nothing virtual about `std::shared_ptr<int>`, only the class that holds it needs virtual serialization...)

lifflander:
The function `allocateConstructForPointer` is supposed to call `SerializeAsVirtualIfNeeded` which checks to see if virtual serialization is needed or not. It looks like this might not be working properly?

cz4rs:
to do:
- [x] fix virtual serialization / bring back `enable_if`s to check that we use `Footprinter` (for raw pointers and std::shared_ptr)
- [ ] change `std::_List_iterator<T>` to `std::list<T>::iterator`

cz4rs:
@lifflander 
~~virtual serialization (`std::shared_ptr`) commit: 8157461~~
uncommenting the `enable_if` in `shared_ptr_serialize.h` triggers the bug, resulting in errors during compilation

~~testcase: `test_virtual_serialize.cc'~~

edit: sorry, to much history rewrites: since this uses regular serialization instead of `getMemoryFootprint`, I will extract this into a separate testcase

cz4rs:
@lifflander 
for the virtual serialization / shared_ptr issue, please see f5f1782

lifflander:
> @lifflander
> for the virtual serialization / shared_ptr issue, please see [f5f1782](https://github.com/DARMA-tasking/checkpoint/commit/f5f178204c168dd3d839db9ee901843e8dee8b77)

@cz4rs Ok, I've solved one problem. Right now, the way virtual serialize works is that it instantiates for all serializer types because it is dynamically dispatched. That is why you are getting the error with Clang when the enable_if is present.

lifflander:
The other problem is that `checkpoint::SerializableTraits<T>` checks for `checkpoint::Serializer` to be dispatched on a intrusive or non-intrusive serialize function. However, when enable_if is on the `checkpoint::Footprinter` it doesn't check for convertibility from `checkpoint::Footprinter` to `checkpoint::Serializer`.

lifflander:
I think I fixed all the bugs (except for list iterator, which I'll leave for you).

- I expanded `SerializableTraits<T>` to `SerializableTraits<T,U>` where `U` is a `SerializerT` so it doesn't assume `checkpoint::Serializer` which leads to convertibility issues.
- For the automatic instantiation in virtual serialize, I first check that the object is actually serializable with the `SerializerT` before trying to instantiate it with the pack of `SerializerTs`
- I put the `std::enable_if` at the top-level of `serialize(SerializerT& s, std::shared_ptr<T>& ptr)` so it can filter out early
- Then, to make the final test compile, I had to also push the enable_if check up into the serializer for `TestDerived2`. The problem is that if `TestDerived2` is serializable with any serializer it assumes that its children will be (I'm still thinking about if there is a way to recursively check that at compile-time). For now, the way to avoid that is to make it's children be serializable or push the check up (which I decided to do in the code).

cz4rs:
@PhilMiller can you please have a look at this? another approval would be good since Jonathan has also added some commits here

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -81,29 +84,23 @@ void serializeRawPtr(SerializerT& s, T* ptr) {
...
-  typename = std::enable_if_t<
-    std::is_same<
-      SerializerT,
-      checkpoint::Footprinter
-    >::value
-  >
+  typename SerializerT
 >
-void serializeFilePtr(SerializerT& s, FILE* ptr) {
+void serializeRawPtr(SerializerT& s, FILE* ptr) {
```
PhilMiller, line 103: In the future, if we want to be very exacting about this, or need to support a broader range of possibly-incomplete types, we could use the `type_if_defined` template described [here](https://devblogs.microsoft.com/oldnewthing/20190712-00/?p=102690)

src/checkpoint/container/vector_serialize.h
```diff
@@ -52,13 +52,23 @@
...
+template <typename SerializerT, typename T, typename VectorAllocator>
+typename std::enable_if_t<
+  std::is_same<SerializerT, checkpoint::Footprinter>::value,
+  void
+> serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) { }
+
+
 template <typename Serializer, typename T, typename VectorAllocator>
 void serialize(Serializer& s, std::vector<T, VectorAllocator>& vec) {
   if (s.isFootprinting()) {
```
PhilMiller: Since `serializeVectorMeta` is now polymorphic on footprinting or not, we could simplify the logic here a bit, to

```
serializeVectorMeta(s, vec);
size_t num_elements = s.isFootprinting() ? vec.capacity() : vec..size();
dispatch::serializeArray(s, &vec[0], num_elements);
```

And have the `s.countBytes(vec)` in the meta function.

src/checkpoint/serializers/footprinter.h
```diff
@@ -61,7 +61,7 @@ struct Footprinter : Serializer {
   }
 
   template<typename T>
```
PhilMiller, line 63: While we're working here, could you also add a simple `void addBytes(size_t s)` method that can be used for containers in the runtime that might keep track of the footprint of objects they're holding, but which can't be traversed directly (e.g. the message queues)

PhilMiller: Body would literally be `num_bytes_ += s;`

PhilMiller:
Get the `Footprinter::addBytes` in, and I'm +1 on this.

cz4rs:
> Get the `Footprinter::addBytes` in, and I'm +1 on this.

`addBytes` added and vector serialization simplified :ok_hand:

PhilMiller (approved):

tests/unit/test_footprinter.cc
```diff
@@ -98,6 +113,15 @@ TEST_F(TestFootprinter, test_fundamental_types) {
     delete ptr;
   }
 
+  {
+    int i = 3;
```
PhilMiller, line 117: @cz4rs I just found that this generates a warning:

```
/Users/pbmille/repos/vt/lib/checkpoint/tests/unit/test_footprinter.cc:117:9: warning: declaration shadows a local variable [-Wshadow]
    int i = 3;
        ^
/Users/pbmille/repos/vt/lib/checkpoint/tests/unit/test_footprinter.cc:80:7: note: previous declaration is here
  int i;
      ^
```

Please open an issue and address it, by using a different name or something

cz4rs: #134 created

PR #131, opened 2020-10-01
130 Expand extent-only up to 4-D; make compile-time fail if not implemented

### LINKED ISSUE #130: Missing extent-only serializers > 2 dimensions
We are missing extent-only serializers for more than 2 dimensions. Also, checkpoint silently fails when the number of dimensions are > 2 due to how the templates are matched (`T = T*`).

lifflander:
Fixes #130

### DIFF
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -522,25 +522,109 @@ inline void serialize(SerializerT& s, Kokkos::DynRankView<T,Args...>& view) {
   SerializeConst<ViewType>::apply(s,view);
 }
 
-template< typename SerializerT, typename T, typename... Ts >
-void serializeExtentOnly(SerializerT& s, Kokkos::View<T*,Ts...>& v, std::string label ) {
-  // Pass label explicitly to reduce network transfer bytes
-  auto view_extent_0 = v.extent(0);
-  s | view_extent_0;
-  if (s.isUnpacking()) {
-    v = Kokkos::View<T*>(label, view_extent_0);
-  }
-}
+template <
+  typename ViewType,
+  typename T = typename ViewGetType<ViewType>::DataType
+>
+struct CountStaticDims {
+  static constexpr int ndims = 0;
+};
 
-template< typename SerializerT, typename T, typename... Ts >
-void serializeExtentOnly(SerializerT& s, Kokkos::View<T**,Ts...>& v, std::string label ) {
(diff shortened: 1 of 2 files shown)


src/checkpoint/container/view_serialize.h
```diff
@@ -522,25 +522,109 @@ inline void serialize(SerializerT& s, Kokkos::DynRankView<T,Args...>& view) {
...
+};
+
+template <typename SerializerT, typename ViewType, typename = void>
+struct SerializeExtentOnlyImpl;
+
+template <typename SerializerT, typename ViewType>
+struct SerializeExtentOnlyImpl<
+  SerializerT,
+  ViewType,
+  typename std::enable_if_t<CountStaticDims<ViewType>::ndims == 1>
```
PhilMiller: Rather than the `enable_if` that will make for lots of expansions etc, why not just specialize on `int N_DIMS`?

src/checkpoint/container/view_serialize.h
```diff
@@ -522,25 +522,109 @@ inline void serialize(SerializerT& s, Kokkos::DynRankView<T,Args...>& view) {
...
+      v = ViewType(
+        label, view_extent_0, view_extent_1, view_extent_2, view_extent_3
+      );
+    }
+  }
+};
+
+template <typename SerializerT, typename ViewType>
+void serializeExtentOnly(SerializerT& s, ViewType& v, std::string label) {
+  SerializeExtentOnlyImpl<SerializerT, ViewType>::apply(s, v, label);
```
PhilMiller: Per the above comment, the `CountStaticDims<ViewType>::ndims` would then get passed as a template argument here.

PhilMiller: If done this way, I think the compiler errors on the case of a missing specialization would probably be somewhat better, too.

PhilMiller:
I'm happy with this now, but having modified it, I don't think I should be the one to provide the sole approval.

lifflander:
> I'm happy with this now, but having modified it, I don't think I should be the one to provide the sole approval.

The changes look good to me

PhilMiller (approved):

PhilMiller:
If we were feeling especially ambitious, we'd find a way to make it work for arbitrary dimensions, with some sort of wild pack expansion and use of `std::iota` or something.

PhilMiller:
Or, we'd request (propose a PR for?) Kokkos to add array-based API for getting extents and initializing new instances

Issue #138, opened 2020-10-08
Virtual serialization: Eliminate need to name parent class in macro arguments


PhilMiller:
Follow-on/companion to #132.

lifflander:
Here's my idea. Say we have some arbitrary classes registered: A, B, C, D and we don't know what relationship they are in. We call `std::is_base_of<X, Y>` for all pairs which tells if `Y` is a derived class of `X`. If the hierarchy is A <- B <- C <- D, 

is_base_of<A, D> = true
is_base_of<A, C> = true
is_base_of<A, B> = true

Thus, `A` is the first base class because all returned true. We remove `A` from the list since it's the first comparison with all true outputs.
Then, we apply recursively starting with `B`.

is_base_of<B, C> = true
is_base_of<B, D> = true

I think the idea of dynamic_cast might also work.

lifflander:
Also, this might be insane but can we do something with sizeof?

PhilMiller:
I like the is_base_of line of thought, and we'd be able to topologically sort on the relationships if we could do it. Problem is, that's template stuff, and so we'd need a compile time list of the classes involved. Can't get that with separate compilation.

I don't think there's anything to be done with sizeof

PhilMiller:
For reference, here's the private message I sent earlier:

I  just had a thought on a really cheesy but seemingly workable way to handle arbitrary single-rooted multiple inheritance (I.e. semi-lattice of types). Have a registry associated with the root. Each class inserts a function into the registry that takes a `BaseT*b`, tries to `dynamic_cast<DerivedT*>(b)`, and calls serialize on that if it succeeds. Then, to serialize such an object, just iterate through the whole registry, calling every entry! This ignores the possibility of ordering dependencies between serializers, but we can get there with some topological sorting or something.

PhilMiller:
My initial suggestion is O(c) in the number of classes in a lattice, for each serialization operation. It would be nice if we instead could memoize by having that function return true/false, and remembering which ones return true. It would cut that down to O(#ancestors), which is optimal.

I don't think we can do that just once at the registry level, though, because we need actual instances in hand on which to call `dynamic_cast`

PhilMiller:
Thinking about this some more, my "some topological sorting" isn't really an answer to the serializer ordering problem. It should not be too hard to run the root before any of the others, but everything else seems much harder.

PhilMiller:
[...]

PR #150, opened 2020-10-26
#138: Virtual Serializer: Automatically determine base classes to call

PhilMiller:


PhilMiller:
Ok, generating and registering the `dynamic_cast` works. Now to write the code to actually call and use it!

PR #152, opened 2020-11-02
#151: Split up virtual serialization logic and better support in-place deserialization

### LINKED ISSUE #151: Handle reconstruction of virtual serialized types sensibly
This will allow users to skip that if they allocate the object for themselves and don't need it.

PhilMiller:


src/checkpoint/dispatch/vrt/virtual_serialize.h
```diff
@@ -157,21 +201,30 @@ struct SerializeAsVirtualIfNeeded<
...
  *   };
  *
- *
  * \param[in] s the serializer
  * \param[in] target a reference to a pointer to the target object
  */
 template <typename SerializerT, typename T>
-void allocateConstructForPointer(SerializerT& s, T*& target) {
-  SerializeAsVirtualIfNeeded<T, SerializerT>::apply(s, target);
+void reconstructPointedToObjectIfNeeded(SerializerT& s, T*& target) {
```
cz4rs, line 216: naming: is `reconstructPointeeIfNeeded` an option?

src/checkpoint/dispatch/dispatch_virtual.h
```diff
@@ -94,7 +107,7 @@
  *           if (!is_null) {
  *             // During size/pack, save the actual derived type of raw_ptr;
  *             // During unpack, allocate/construct raw_ptr with correct virtual type
- *             checkpoint::allocateConstructForPointer(s, raw_ptr);
+ *             checkpoint::reconstructPointedToObjectIfNeeded(s, raw_ptr);
```
cz4rs, line 110: have you checked if this snippet (serializing raw pointer) actually works? IIRC the last time I've tried it with `allocateConstructForPointer` it would not compile

cz4rs (approved):
looks good, only minor nitpicking in comments

src/checkpoint/dispatch/vrt/virtual_serialize.h
PhilMiller: I'd worry about it being misread as Pointer. I'll admit this is way more verbose than I'd like.

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: I'll give it a try in the tests.

cz4rs: I have checked the old version and this is working just fine:
```
template <typename T>
struct MyObjectWithRawPointer {
  T* raw_ptr;

  template <typename SerializerT>
  void serialize(SerializerT& s) {
    bool is_null = raw_ptr == nullptr;
    s | is_null; // serialize whether we have a null pointer not
    if (!is_null) {
      // During size/pack, save the actual derived type of raw_ptr;
      // During unpack, allocate/construct raw_ptr with correct virtual type
      checkpoint::allocateConstructForPointer(s, raw_ptr);
      s | *raw_ptr;
    }
  }
};

TEST_F(SomeTest, test_raw_pointer) {
  MyObjectWithRawPointer<int> obj;
  obj.raw_ptr = nullptr;

  auto ret_null = checkpoint::serialize(obj);

  int i = 3;
  obj.raw_ptr = &i;

  auto ret = checkpoint::serialize(obj);
}
```
so maybe I was just imagining things, but anyways, a test (or example?) like this could be good to have

src/checkpoint/dispatch/vrt/virtual_serialize.h
cz4rs: ok, explicit is probably better than concise in the interface

src/checkpoint/dispatch/dispatch_virtual.h
PhilMiller: I just added such an example. All good?

PR #157, opened 2020-11-19
156: footprint types that are not serializable

### LINKED ISSUE #156: footprint types that are not serializable
provide an estimated footprint for containers whose elements are not serializable (eg. `std::vector<MPI_Request>`)

proposed implementation is 
```cpp
s.countBytes(container);
s.addBytes(container.size() * sizeof(T));
```

cz4rs:
fixes #156 

current pain points in `vt`:
- [x] Open MPI types

- [x] `std::list` iterators

note:
- [ ] general solution for any non-serializable type in a container would be best


PhilMiller:
I'd really like to see that `countBytes` logic applied to any non-serializable type in a container. Did you see my suggestion for a potentially appropriate type trait? It would go on the `enable_if` for the footprinting-specific container `serialize` functions

cz4rs:
> I'd really like to see that `countBytes` logic applied to any non-serializable type in a container. Did you see my suggestion for a potentially appropriate type trait? It would go on the `enable_if` for the footprinting-specific container `serialize` functions

Yes, I have seen it, but I was unable to get it to work so far (various errors at the dispatch level). Let me push a cleaned up example for easier discussion.

cz4rs:
ed726eb and b41585c are two main lines that I've tried, but they are both not working - resulting in the same error as with the serializer completely missing

src/checkpoint/container/vector_serialize.h
```diff
@@ -100,6 +100,22 @@ void serialize(Serializer& s, std::vector<bool, VectorAllocator>& vec) {
...
+  typename SerializerT,
+  typename T,
+  typename VectorAllocator,
+  typename = std::enable_if_t<
+    std::is_same< // is_footprintable
+      size_t,
+      decltype(checkpoint::getMemoryFootprint(std::declval<T>()))
+    >::value
+  >
+>
```
PhilMiller: OK, looking at the rest of this file's code, I think it should maybe first be applied to specializations of `dispatch::serializeArray` instead of `serialize(vector<T>)`. Also, I think the sense of the condition is reversed - we want to consider `sizeof(T)` when `!is_footprintable<T>`

tests/unit/test_footprinter.cc
```diff
@@ -543,4 +555,61 @@ TEST_F(TestFootprinter, test_virtual_serialize) {
...
+  int i;
+};
+
+TEST_F(TestFootprinter, test_no_serialize) {
+  std::vector<TestNoSerialize> v(7);
+
+  std::cout << std::boolalpha;
+  std::cout << SerializableTraits<int>::is_footprintable_v << std::endl;
+  std::cout << SerializableTraits<Test1>::is_footprintable_v << std::endl;
+  std::cout << SerializableTraits<TestNoSerialize>::is_footprintable_v << std::endl;
```
cz4rs: no success with the generalized approach, I've just pushed a somewhat cleaned up current state and I'm throwing in the towel for today

- I have moved the detection into `dispatch::serializeArray`
- `is_footprintable` in current form doesn't work as intended - here, calling it on `TestNoSerialize` class yields `true`
- I have tried to use somewhat along the lines of
```cpp
not (SerializableTraits<T>::has_serialize_function or
  SerializableTraits<T>::is_bytecopyable)
```
but couldn't get it to work either (dealing with redefinition, ambiguous call, template deduction errors)

I have implemented serializer for `std::list::iterator` in the meantime, but this came with its own issues (see comment in `list_serialize.h`).

cz4rs: @PhilMiller @lifflander I would appreciate any help with creating the general solution (a footprinting serializer for containers whose elements are not serializable, in *vt* we have such case with `std::vector` and `std::unordered_map`)

PhilMiller:
Ok, I've gotten the type trait stuff working for the `serializeArray` case. However, I realize that more sensible thing to do is maybe solve this at a slightly lower level, in the `Traverse` machinery or below, so that it works for containers other than `std::vector` more automatically.

src/checkpoint/dispatch/dispatch_serializer_nonbyte.h
```diff
@@ -161,6 +170,23 @@ struct SerializerDispatchNonByte {
...
+    SerializerT& s, T* val, SerialSizeType num, justFootprint<U>* = nullptr
+  ) {
+    debug_checkpoint(
+      "SerializerDispatch: justFootprint: val=%p\n",
+      static_cast<void*>(&val)
+    );
+    static bool firstCall = true;
+    if (firstCall) {
+      firstCall = false;
+      // optionally print a warning here
```
PhilMiller: We need to decide what to do here.

PhilMiller:
I just tested that this passes all the footprinting tests in vt.

cz4rs:
@PhilMiller this looks good to me, I can see that the PR check is complaining about signatures missing (in your commits) - feel free to force push the amendments

@lifflander this probably requires you review as a third party

edit: after giving it a second thought, this probably needs an explicit mention in the documentation, because now user will always be able to footprint stuff, but might be unaware about a silent fallback to this approach

src/checkpoint/dispatch/dispatch_serializer_nonbyte.h
cz4rs: hard to say how much additional value a separate warning would provide, maybe expanding the first (unconditional) print on the entrance would be sufficient (e.g. adding `typeid`)?

PhilMiller: My thinking is that we may want a printed warning here in much broader circumstances than when we have full debug printing turned on. Perhaps aligned with `CMAKE_BUILD_TYPE=Debug` or such.

PhilMiller: Regardless, both this debug print and the ones below should probably have a name from `typeid` incorporated.

lifflander: I agree that printing `typeid` would be good here.

src/checkpoint/dispatch/dispatch_serializer_nonbyte.h
```diff
@@ -131,6 +131,15 @@ struct SerializerDispatchNonByte {
...
     typename std::enable_if<std::is_enum<U>::value, T>::type;
+
+    template <typename U>
+    using justFootprint =
+    typename std::enable_if<
+      std::is_same<S, checkpoint::Footprinter>::value and
+      not SerializableTraits<U, S>::is_traversable and
+      not std::is_enum<U>::value,
+      T
+      >::type;
```
lifflander: Slight indentation problem here on `>::type;` line

src/checkpoint/container/raw_ptr_serialize.h
```diff
@@ -93,17 +93,30 @@ void serializeRawPtr(SerializerT& s, void* ptr) {
   s.countBytes(ptr);
 }
 
+#define CHECKPOINT_FOOTPRINT_PIMPL_WITH_SIZEOF_PTR(PIMPL_TYPE) \
+  template < \
+             typename SerializerT, \
```
lifflander: Fix indentation here?

lifflander (approved):
Overall this looks good. A few minor points about formatting fixes

PR #160, opened 2020-11-19
158: fix std::vector serialization

### LINKED ISSUE #158: footprinting: do not traverse spare memory past the last element of std::vector
Consider an array element type that has ‘interesting’ members whose serialization might involve dereferencing a pointer. Because we look at the capacity rather than the length when footprinting, we’ll try to call the serialize methods on the spare memory past the end of the array of live elements. 
https://github.com/DARMA-tasking/checkpoint/blob/53f5c27816a3fe099c6c19b643a220c1d2d19485/src/checkpoint/container/vector_serialize.h#L77-L78
Vector's serializer should always call `dispatch::serializeArray(s, &vec[0], vec.size());` and then `s.addBytes(sizeof(T) * (vec.capacity() - vec.size()));` when footprinting.

### LINKED ISSUE #159: segmentation fault when serializing empty vector using virtual serialize
Following code:
```cpp
  std::vector<TestDerived2> v;
  checkpoint::getMemoryFootprint(v);
```
where `TestDerived2` is:
https://github.com/DARMA-tasking/checkpoint/blob/53f5c27816a3fe099c6c19b643a220c1d2d19485/tests/unit/test_footprinter.cc#L494-L517

triggers segmentation fault with following backtrace:
```gdb
#0  0x00005555555987d2 in checkpoint::dispatch::vrt::virtualSerialize<checkpoint::tests::unit::TestDerived2, checkpoint::Footprinter> (base=@0x7fffffffd418: 0x0, s=...)
    at ../src/checkpoint/dispatch/vrt/virtual_serialize.h:66
#1  0x00005555555963f0 in checkpoint::dispatch::SerializerDispatchNonByte<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2, checkpoint::dispatch::BasicDispatcher<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2> >::apply<checkpoint::tests::unit::TestDerived2> (this=0x7fffffffd4a7, s=..., val=0x0, num=0)
    at ../src/checkpoint/dispatch/dispatch_serializer_nonbyte.h:212
#2  0x000055555559447c in checkpoint::dispatch::SerializerDispatchNonByte<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2, checkpoint::dispatch::BasicDispatcher<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2> >::operator()<checkpoint::tests::unit::TestDerived2> (this=0x7fffffffd4a7, s=..., val=0x0, num=0)
    at ../src/checkpoint/dispatch/dispatch_serializer_nonbyte.h:101
#3  0x0000555555592067 in checkpoint::dispatch::SerializerDispatchByte<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2, checkpoint::dispatch::BasicDispatcher<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2> >::operator()<checkpoint::tests::unit::TestDerived2>(checkpoint::Footprinter&, checkpoint::tests::unit::TestDerived2*, unsigned long, std::enable_if<!checkpoint::SerializableTraits<checkpoint::tests::unit::TestDerived2, void>::is_bytecopyable, checkpoint::tests::unit::TestDerived2>::type*) (this=0x7fffffffd4ef, s=..., val=0x0, num=0, 
    x=0x0) at ../src/checkpoint/dispatch/dispatch_serializer_byte.h:107
#4  0x000055555558e9a9 in checkpoint::dispatch::Traverse::with<checkpoint::tests::unit::TestDerived2, checkpoint::Footprinter> (target=..., t=..., len=0)
    at ../src/checkpoint/dispatch/dispatch.impl.h:71
#5  0x0000555555597a85 in checkpoint::dispatch::serializeArray<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2> (s=..., array=0x0, len=0)
    at ../src/checkpoint/dispatch/dispatch.impl.h:125
#6  0x00005555555957b6 in checkpoint::serialize<checkpoint::Footprinter, checkpoint::tests::unit::TestDerived2, std::allocator<checkpoint::tests::unit::TestDerived2> > (s=..., 
    vec=std::vector of length 0, capacity 0) at ../src/checkpoint/container/vector_serialize.h:78
(...)
```

Please note that serializing empty vector of class without virtual serialize seems to run fine (although it could be innocently looking UB).

Major suspect:
https://github.com/DARMA-tasking/checkpoint/blob/53f5c27816a3fe099c6c19b643a220c1d2d19485/src/checkpoint/container/vector_serialize.h#L78
into:
https://github.com/DARMA-tasking/checkpoint/blob/53f5c27816a3fe099c6c19b643a220c1d2d19485/src/checkpoint/dispatch/dispatch.impl.h#L123-L126

cz4rs:
fixes #158 
fixes #159 

- do not traverse spare memory past the last element of vector (but account for it when footprinting)
- use data() to guarantee valid pointer to underlying array (even for empty container)
- avoid dereferencing null pointers on dispatch
- add adequate unit test

### DIFF
--- a/src/checkpoint/dispatch/dispatch.impl.h
+++ b/src/checkpoint/dispatch/dispatch.impl.h
@@ -122,7 +122,7 @@ T* Standard::construct(SerialByteType* mem) {
 
 template <typename Serializer, typename T>
 inline void serializeArray(Serializer& s, T* array, SerialSizeType const len) {
-  if (array) {
+  if (len > 0) {
     Traverse::with<T, Serializer>(*array, s, len);
   }
 }
--- a/src/checkpoint/dispatch/dispatch.impl.h
+++ b/src/checkpoint/dispatch/dispatch.impl.h
@@ -122,7 +122,9 @@ T* Standard::construct(SerialByteType* mem) {
 
 template <typename Serializer, typename T>
 inline void serializeArray(Serializer& s, T* array, SerialSizeType const len) {
-  Traverse::with<T, Serializer>(*array, s, len);
+  if (array) {
+    Traverse::with<T, Serializer>(*array, s, len);
+  }
 }
 
 template <typename T>
--- a/tests/unit/test_footprinter.cc
+++ b/tests/unit/test_footprinter.cc
@@ -76,6 +76,47 @@ void serialize(Serializer& s, Test2 t) {
   s | t.f;
 }
 
+struct TestBase {
+
+  checkpoint_virtual_serialize_root()
+
+  virtual ~TestBase() = default;
+
+  template <typename SerializerT>
+  void serialize(SerializerT& s) {
+    s | a;
+  }
+
+private:
+  int a = 0;
+};
+
+struct TestDerived2 : TestBase {
+  explicit TestDerived2(int i) {}
+  explicit TestDerived2(SERIALIZE_CONSTRUCT_TAG) {}
+
+  checkpoint_virtual_serialize_derived_from(TestBase)
+
+  template <
+    typename SerializerT,
+    typename = std::enable_if_t<
+      std::is_same<
+        SerializerT,
+        checkpoint::Footprinter
+      >::value
+    >
+  >
+  void serialize(SerializerT& s) {
+    s | raw_pointer;
+    s | shared_pointer;
+  }
+
+private:
+  std::shared_ptr<int> shared_pointer = std::make_shared<int>(5);
+  int* raw_pointer = nullptr;
+};
+
+
 TEST_F(TestFootprinter, test_fundamental_types) {
   int i;
   std::size_t size = 7;
@@ -474,48 +515,15 @@ TEST_F(TestFootprinter, test_vector) {
       sizeof(v)
     );
   }
+
+  {
+    std::vector<TestDerived2> v;
+    EXPECT_EQ(
+      checkpoint::getMemoryFootprint(v),
+      sizeof(v)
+    );
+  }
 }
-
-struct TestBase {
-
-  checkpoint_virtual_serialize_root()
-
-  virtual ~TestBase() = default;
-
-  template <typename SerializerT>
-  void serialize(SerializerT& s) {
-    s | a;
-  }
-
-private:
-  int a = 0;
-};
-
-struct TestDerived2 : TestBase {
-  explicit TestDerived2(int i) {}
-  explicit TestDerived2(SERIALIZE_CONSTRUCT_TAG) {}
-
-  checkpoint_virtual_serialize_derived_from(TestBase)
-
-  template <
-    typename SerializerT,
-    typename = std::enable_if_t<
-      std::is_same<
-        SerializerT,
-        checkpoint::Footprinter
-      >::value
-    >
-  >
-  void serialize(SerializerT& s) {
-    s | raw_pointer;
-    s | shared_pointer;
-  }
-
-private:
-  std::shared_ptr<int> shared_pointer = std::make_shared<int>(5);
-  int* raw_pointer = nullptr;
-};
-
 
 TEST_F(TestFootprinter, test_virtual_serialize) {
   {
--- a/tests/unit/test_footprinter.cc
+++ b/tests/unit/test_footprinter.cc
@@ -518,9 +518,15 @@ TEST_F(TestFootprinter, test_vector) {
 
   {
     std::vector<TestDerived2> v;
-    EXPECT_EQ(
+    EXPECT_GE(
       checkpoint::getMemoryFootprint(v),
       sizeof(v)
+    );
+
+    v.reserve(3);
+    EXPECT_EQ(
+      checkpoint::getMemoryFootprint(v),
+      sizeof(v) + v.capacity() * sizeof(TestDerived2)
     );
   }
 }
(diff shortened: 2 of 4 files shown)


src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -122,7 +122,9 @@ T* Standard::construct(SerialByteType* mem) {
 
 template <typename Serializer, typename T>
 inline void serializeArray(Serializer& s, T* array, SerialSizeType const len) {
-  Traverse::with<T, Serializer>(*array, s, len);
+  if (array) {
```
cz4rs: I'm not sure if this is enough:

> https://en.cppreference.com/w/cpp/container/vector/data
If size() is 0, data() may or may not return a null pointer.

PhilMiller: How about `if (len > 0)`?

tests/unit/test_footprinter.cc
```diff
@@ -474,48 +515,15 @@ TEST_F(TestFootprinter, test_vector) {
...
-        checkpoint::Footprinter
-      >::value
-    >
-  >
-  void serialize(SerializerT& s) {
-    s | raw_pointer;
-    s | shared_pointer;
+  {
+    std::vector<TestDerived2> v;
+    EXPECT_EQ(
```
PhilMiller: Could this potentially return greater, if the default-constructed vector has a non-zero capacity?

cz4rs: ~I was just testing that case (ad53ede) - it was too early to celebrate victory~

cz4rs: yes, this is implementation dependent, so I got lucky - changed to `GE`

src/checkpoint/dispatch/dispatch.impl.h
cz4rs: ad53ede crashed with non-null pointer (size = 0, capacity > 0), but `if (len > 0)` seems to work

tests/unit/test_footprinter.cc
```diff
@@ -474,48 +515,21 @@ TEST_F(TestFootprinter, test_vector) {
...
-  explicit TestDerived2(int i) {}
-  explicit TestDerived2(SERIALIZE_CONSTRUCT_TAG) {}
 
-  checkpoint_virtual_serialize_derived_from(TestBase)
+  {
+    std::vector<TestDerived2> v;
+    EXPECT_GE(
+      checkpoint::getMemoryFootprint(v),
+      sizeof(v)
+    );
```
PhilMiller, line 524: I think this test could still fail spuriously on some platforms, if a default-constructed `vector<T> v` has `v.capacity() > 0`.

cz4rs: `EXPECT_GE` (greater or equal) should cover this

PhilMiller: Oops, sorry, I misread that this says `EXPECT_GE` now.

tests/unit/test_footprinter.cc
```diff
@@ -474,48 +515,21 @@ TEST_F(TestFootprinter, test_vector) {
...
-        SerializerT,
-        checkpoint::Footprinter
-      >::value
-    >
-  >
-  void serialize(SerializerT& s) {
-    s | raw_pointer;
-    s | shared_pointer;
+    v.reserve(3);
+    EXPECT_EQ(
```
PhilMiller, line 527: I think this case also needs to be `EXPECT_GE` though - there's no requirement that the memory allocated by `reserve` be a 'tight fit'

tests/unit/test_footprinter.cc
```diff
@@ -474,48 +515,21 @@ TEST_F(TestFootprinter, test_vector) {
...
+      checkpoint::getMemoryFootprint(v),
+      sizeof(v) + v.capacity() * sizeof(TestDerived2)
```
cz4rs, line 529: ~ok, to avoid surprises, this could also fail on some platforms (if capacity was bigger than 3 by default):~
> reserve() cannot be used to reduce the capacity of the container; to that end shrink_to_fit() is provided. 

~`shrink_to_fit()` incoming~

tests/unit/test_footprinter.cc
cz4rs: you beat me by 90 seconds with this comment :)

lifflander (approved):
This look solid to me

tests/unit/test_footprinter.cc
PhilMiller: It's not just if capacity is greater than 3 by default, it's also if the `reserve` call gives back a capacity greater than 3. This could happen if it rounds to some unit that it knows the allocator works in, for instance, or if the library implements http://wg21.link/p0401

tests/unit/test_footprinter.cc
cz4rs: in the end, it will always work because the comparison uses `v.capacity()` (only `reserve` uses magic number)

PhilMiller: I see, great.

PhilMiller (approved):

PR #164, opened 2020-11-24
#163: build: correct messages when package not found

### LINKED ISSUE #163: Bugs in find_package_local
When `find_package_local` fails to find the package, the list of search paths that it prints is incorrect.  The printing logic assumes that the `${pkg_name}_DIR` variable was unchanged during the loop over potential subdirectories containing the needed cmake files, which is incorrect.  Additionally, the subdirectory existence check includes an undefined variable, `pkg_prefix`.

nlslatt:
Fixes #163.

This prints the paths it actually checks, and doesn't break anything for the two use cases I tested.


lifflander (approved):
Looks correct to me now!

PR #166, opened 2020-11-25
162: preserve std::vector and std::unordered_map capacity

### LINKED ISSUE #162: Serialization of vector and unordered_map does not preserve capacity
A code may have intentionally enlarged these containers to avoid potential pitfalls from reallocation. In the vector case, that pitfall may even change correctness and not just performance, if it set a sufficient size to avoid pointer/iterator invalidation at some other point during execution.

cz4rs:
fixes #162

### DIFF
--- a/src/checkpoint/container/vector_serialize.h
+++ b/src/checkpoint/container/vector_serialize.h
@@ -57,6 +57,10 @@ typename std::enable_if_t<
   not std::is_same<SerializerT, checkpoint::Footprinter>::value,
   void
 > serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) {
+  SerialSizeType vec_capacity = vec.capacity();
+  s | vec_capacity;
+  vec.reserve(vec_capacity);
+
   SerialSizeType vec_size = vec.size();
   s | vec_size;
   vec.resize(vec_size);
@@ -69,7 +73,6 @@ typename std::enable_if_t<
 > serializeVectorMeta(SerializerT& s, std::vector<T, VectorAllocator>& vec) {
   s.countBytes(vec);
 }
-
 
 template <typename Serializer, typename T, typename VectorAllocator>
 void serialize(Serializer& s, std::vector<T, VectorAllocator>& vec) {
--- a/tests/unit/test_container.cc
+++ b/tests/unit/test_container.cc
@@ -181,7 +181,7 @@ TYPED_TEST_CASE_P(TestMultiContainer);
 TYPED_TEST_CASE_P(TestMultiContainerUnordered);
 
 template <typename ContainerT, typename Pair>
-static void testMultiContainer(bool is_ordered, std::initializer_list<Pair> lst) {
+static void testMultiContainerOrdered(std::initializer_list<Pair> lst) {
   ContainerT c1{lst};
   auto ret = checkpoint::serialize(c1);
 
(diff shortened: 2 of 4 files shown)


cz4rs:
note: `std::unordered_multimap` should probably get the same treatment

src/checkpoint/container/vector_serialize.h
```diff
@@ -60,6 +60,10 @@ typename std::enable_if_t<
   SerialSizeType vec_size = vec.size();
   s | vec_size;
   vec.resize(vec_size);
+
+  SerialSizeType vec_capacity = vec.capacity();
+  s | vec_capacity;
+  vec.reserve(vec_capacity);
```
PhilMiller, line 62: This should be done before resizing, so that it doesn't reallocate twice.

tests/unit/test_container.cc
```diff
@@ -97,6 +97,22 @@ static void testContainer(bool is_ordered, std::initializer_list<T> lst) {
...
+  bool /*is_ordered*/, std::initializer_list<int> lst
+) {
+  std::vector<int> c1{lst};
+  c1.reserve(1000);
+  auto ret = checkpoint::serialize(c1);
+
+  auto t1 = checkpoint::deserialize<std::vector<int>>(ret->getBuffer());
+
+  EXPECT_EQ(c1.size(), t1->size());
+  EXPECT_EQ(c1.capacity(), t1->capacity());
```
PhilMiller: The reserved capacity should be a named constant, and the comparison should be that the deserialized vector has capacity greater than or equal to both the constant and the original vector. The call to `reserve` can over allocate at both sites

PhilMiller:
The vector code looks good and as expected.

tests/unit/test_container.cc
cz4rs: fixed, explanation added in the comment

PhilMiller:
Do the hash table types also need to do something to preserve maximum load factor or other parameters?

cz4rs:
> Do the hash table types also need to do something to preserve maximum load factor or other parameters?

good catch with max load factor, it should be (de)serialized before rehashing
other than that, I don't see anything in unordered_map interface that could be preserved

edit: changes for maximum load factor + `std::unordered_set` and `std::unordered_multiset` incoming

lifflander (approved):
Looks good to me

tests/unit/test_container.cc
```diff
@@ -82,28 +82,62 @@ static void testEqualityContainerUnordered(ContainerT& c1, ContainerT& t1) {
...
+  std::size_t const custom_bucket_count = 1000;
+  c1.rehash(custom_bucket_count);
+
+  auto ret = checkpoint::serialize(c1);
+  auto t1 = checkpoint::deserialize<ContainerT>(ret->getBuffer());
+
+  EXPECT_EQ(c1.size(), t1->size());
+  EXPECT_EQ(t1->max_load_factor(), custom_max_load_factor);
+  EXPECT_EQ(t1->max_load_factor(), c1.max_load_factor());
+  EXPECT_GE(t1->bucket_count(), custom_bucket_count);
```
cz4rs, line 129: I wasn't sure about `greater or equal` here (considering guaranteed low load factor), but test fails using `EXPECT_EQ`

cz4rs:
Thanks Jonathan! @PhilMiller if you have no more comments, I will merge this shortly

PhilMiller (approved):

PR #169, opened 2020-11-26
168 CMake: fix bugs in export and installation

### LINKED ISSUE #168: Fix cmake install bug in checkpoint

lifflander:
Fixes #168


nmm0 (approved):
Looks good -- it won't export on non-subdir builds but the `find_dependency` in checkpointConfig.cmake.in should do that job

PR #171, opened 2020-12-02
170: build: work around old cmake (<3.13)

### LINKED ISSUE #170: Check for cmake version to work around cmake < 3.13
Don't install subdirectory target in that case.

lifflander:
Fixes #170

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -144,8 +144,16 @@ if("${CMAKE_SOURCE_DIR}/src" STREQUAL PROJECT_SOURCE_DIR)
   message(STATUS "Checkpoint: building as top-level")
 else()
   message(STATUS "Checkpoint: building as a sub-directory")
-  install(TARGETS detector EXPORT ${CHECKPOINT_LIBRARY})
-  set(DETECTOR_EXPORT detector)
+
+  if(CMAKE_VERSION VERSION_LESS "3.13.0")
+    message(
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -144,8 +144,16 @@ if("${CMAKE_SOURCE_DIR}/src" STREQUAL PROJECT_SOURCE_DIR)
   message(STATUS "Checkpoint: building as top-level")
 else()
   message(STATUS "Checkpoint: building as a sub-directory")
-  install(TARGETS detector EXPORT ${CHECKPOINT_LIBRARY})
-  set(DETECTOR_EXPORT detector)
+
+  if(CMAKE_VERSION VERSION_LESS "3.13.0")
+    message(
+      WARNING "Checkpoint: Install targets skipped, effects packages built with a Checkpoint install"
```
PhilMiller: This warning should probably include a statement of the CMake version dependence and an issue reference in it

cz4rs: typo (?)
```suggestion
      WARNING "Checkpoint: Install targets skipped, affects packages built with a Checkpoint install"
```

cz4rs: ok, it's the same as in #1165

lifflander: Fixed

lifflander: Good catch, fixed!

PR #181, opened 2021-03-03
180: std::unordered_map serializer causing huge memory explosion with empty container

### LINKED ISSUE #180: std::unordered_map serializer causing huge memory explosion with empty container
This code seems like it has a bug that is causing a huge number of unordered_map nodes to be allocated even though the target container is actually empty the entire time. It's causing a massive growth of hundreds of megabytes:

```C++
template <typename SerializerT, typename ContainerT>
inline void serializeUnorderedAssociativeContainer(
  SerializerT& s, ContainerT& cont
) {
  if (not s.isFootprinting()) {
    auto max_load_factor = cont.max_load_factor();
    s | max_load_factor;
    cont.max_load_factor(max_load_factor);

    auto bucket_count = cont.bucket_count();
    s | bucket_count;
    cont.rehash(bucket_count);
  }

  serializeMapLikeContainer(s, cont);
}
```

At the very least we should not be rehashing everything during sizing and packing! That has linear complexity.

Screen Shot 2021-03-03 at 2 25 31 PM
Screen Shot 2021-03-03 at 2 25 46 PM

cz4rs:
fixes #180

### DIFF
--- a/src/checkpoint/container/map_serialize.h
+++ b/src/checkpoint/container/map_serialize.h
@@ -96,7 +96,9 @@ inline void serializeUnorderedAssociativeContainer(
 
     auto bucket_count = cont.bucket_count();
     s | bucket_count;
-    cont.rehash(bucket_count);
+    if (s.isUnpacking() and bucket_count > cont.bucket_count()) {
+      cont.rehash(bucket_count);
+    }
   }
 
   serializeMapLikeContainer(s, cont);
--- a/tests/unit/test_container.cc
+++ b/tests/unit/test_container.cc
@@ -137,6 +137,9 @@ TYPED_TEST_P(TestContainer, test_single_ordered_container) {
 
   using ContainerT = TypeParam;
   using ValueT = typename ContainerT::value_type;
+
+  testContainerOrdered<ContainerT, ValueT>({});
+
   testContainerOrdered<ContainerT, ValueT>(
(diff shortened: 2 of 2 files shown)


cz4rs:
note: the existing test is more or less at:
https://github.com/DARMA-tasking/checkpoint/blob/60d6542a6a760e1865f90780f33a46950e19aa82/tests/unit/test_container.cc#L302-L305

working on checking the proposed fix and improving the test

cz4rs:
with a simple test like this:
```cpp
TEST_F(TestUnorderedMap, test_memory_explosion) {
  for (int i =0; i < 100000; i++) {
    std::unordered_map<std::string, double> c1 = {};
    auto ret = checkpoint::serialize(c1);
    checkpoint::deserialize<std::unordered_map<std::string, double>>(
      ret->getBuffer()
  );
  }
}
```

- `develop`:
image

- d0fac03:
image

Note: the test above gives some idea about what's going on with empty unordered maps, but id **does not reproduce** the explosion in consumption graph (as in the first graph posted by Jonathan).
image

cz4rs:
TODO:
- [ ] work on the unit test  (bc0cecc makes sure that empty containers are tested at all, but does not check explicitly for superfluous `rehash`calls)
- [x] ~~consider the case of `max_load_factor` being modified, but no `rehash` called~~

lifflander (approved):
Looks good.. .this fixes the memory blow-up

PR #189, opened 2021-03-29
188 eti optimize

### LINKED ISSUE #188: Add vaiants of macros for explicit instantiation
For optimizing build times.

lifflander:
Fixes #188 

Quick hack to see how it performs. Lots of copy-paste code that needs to be eliminated.

### DIFF
--- a/src/checkpoint/dispatch/vrt/derived.h
+++ b/src/checkpoint/dispatch/vrt/derived.h
@@ -86,6 +86,49 @@
     return ::checkpoint::dispatch::vrt::objregistry::makeObjIdx<_CheckpointDerivedType>();  \
   }
 
+#define checkpoint_virtual_serialize_derived_decl()                                         \
+  void _checkpointDynamicSerialize(                                                         \
+    void* s,                                                                                \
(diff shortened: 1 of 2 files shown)


src/checkpoint/dispatch/vrt/derived.h
```diff
@@ -86,6 +86,51 @@
...
 
+#define checkpoint_virtual_serialize_derived_decl()                                         \
+  void _checkpointDynamicSerialize(                                                         \
+    void* s,                                                                                \
+    ::checkpoint::dispatch::vrt::TypeIdx base_ser_idx,                                      \
+    ::checkpoint::dispatch::vrt::TypeIdx expected_idx                                       \
+  ) override;                                                                               \
+  ::checkpoint::dispatch::vrt::TypeIdx _checkpointDynamicTypeIndex() override;
+
+#define checkpoint_virtual_serialize_derived_from_def(CLASS, PARENT)                        \
```
lifflander, line 97: I need variants with and without `template <>` here and probably above also.

PR #191, opened 2021-04-13
174 check class correctness

### LINKED ISSUE #174: Add typeid name/type registry to pack/unpack to confirm class correctness
@jstrzebonski Overall, the goal is to improve error checking in checkpoint. This issue is to register class IDs with the type register when we serialize to check if the same class is being traversed in the same order when de-serializing. Right now, we only register types for polymorphic classes because we need to reconstruct them based on runtime type information. In an error checking mode, we want to extend this registration to all classes and then put this information into the serialized buffer so we can check if the traversal matches the same types during de-serialization.

jstrzebonski:
Fixes #174 

TODO:
- [x] depending on error checking mode (ON/OFF) serialize types hashes and names recursively, or not
- [x] could the error print whole serialization path, not only the last wrong type?
- [x] move type_registry out of vrt namespace to a better place
- [x] set error checking flag in dockerfiles and ci/build_cpp.sh
- [x] clean up the code
- [x] add proper unit test


jstrzebonski:
The error is not the prettiest, but I think is readable anyway:
"Unpacking wrong type, got=N10checkpoint5tests4unit11FailureObj1E, expected=N10checkpoint5tests4unit11FailureObj3E"

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -59,23 +63,110 @@ inline Serializer& operator|(Serializer& s, T& target) {
...
 
   SerializerDispatch<TraverserT, CleanT, DispatchType> ap;
-  ap(t, val, len);
+  try {
+    ap(t, val, len);
+  } catch (std::runtime_error &err) {
+    auto const what = std::string{err.what()} +
+                      "\npath: " + vrt::typeregistry::getTypeName<T>();
+    throw std::runtime_error(what);
+  }
```
jstrzebonski: There's a lot of temporary code at the moment, but my general proposition (and IMHO the easiest one) is to simply throw from `withTypeIdx<T>()`, do a `try-catch` around `ap()` and for each catch add information about current type.

I added a temporary super simple test (`test_struct_ee`) to check this, and the result looks like this:

```
Unpacking wrong type, got=i, expected=i
path: N10checkpoint5tests4unit2AAE
path: N10checkpoint5tests4unit2BBE
path: N10checkpoint5tests4unit2CCE
path: N10checkpoint5tests4unit2DDE
path: N10checkpoint5tests4unit2EEE
```

What's left is to actually do this only if error checking is enabled. If it's not, `ap()` needs to be called without `try-catch` (and also whole `withTypeIdx()` should probably be erased).

Does it make any sense @lifflander @PhilMiller ?

lifflander: Yes, this makes sense to me. Although I'm not sure I completely understand the necessity of `DoNothing`. Can you explain why you are checking for an `int` there?

jstrzebonski: That's just temporary code to easily test that try-catch solution. I'll remove it soon.

jstrzebonski:
With demangling added, the exception message is even better:
```
Unpacking wrong type, got=int, expected=int
path: checkpoint::tests::unit::AA
path: checkpoint::tests::unit::BB
path: checkpoint::tests::unit::CC
path: checkpoint::tests::unit::DD
path: checkpoint::tests::unit::EE
```

(And that `got=int, expected=int` is wrong obviously, but it's just for testing purpose).

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -59,23 +64,125 @@ inline Serializer& operator|(Serializer& s, T& target) {
...
+        thisTypeIdx != serTypeIdx) {
+      auto err = std::string("Unpacking wrong type, got=") +
+                 vrt::typeregistry::getTypeNameForIdx(thisTypeIdx) +
+                 " idx=" + std::to_string(thisTypeIdx) + ", expected=" +
+                 vrt::typeregistry::getTypeNameForIdx(serTypeIdx) +
+                 " idx=" + std::to_string(serTypeIdx);
+      throw std::runtime_error(err);
+    }
+  }
+#endif
```
jstrzebonski: That's the very basic check that only checks if `checkpoint::deserialize<T>(buffer)` is called for correct `T`. 
And there is no recursion for this check.

We wanted it to be enabled if "deep" checking is turned off.

.gitignore
```diff
@@ -1,5 +1,6 @@
 build/
 CMakeFiles/
+.vscode/
```
PhilMiller: This sort of thing should probably be in one's own global git ignore configuration file, not in individual repositories.

jstrzebonski: OK, removed.

jstrzebonski: Although I guess that applies to `.dir-locals.el` as well. And if you take a look at vt's .gitignore there might be even more things to clean up :- )

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -60,22 +65,75 @@ inline Serializer& operator|(Serializer& s, T& target) {
...
+  withTypeIdx<CleanT>(t);
+  #endif
 
+  auto val = cleanType(&target);
   SerializerDispatch<TraverserT, CleanT, DispatchType> ap;
+
+  #if defined(SERIALIZATION_ERROR_CHECKING)
+  try {
+    ap(t, val, len);
+  } catch (std::runtime_error& err) {
```
PhilMiller: If the code is going to throw and catch for this, it should probably use a distinct exception type, so that we don't inadvertently catch some other unrelated exception that happens to arise e.g. in a user-written `serialize()` method.

PhilMiller: With a custom exception type, the `path` lines should also include an incrementing number indicating the effective call depth.

PhilMiller:
I like the logic. With some improvement to the error reporting, i think this will be ready to go.

src/checkpoint/dispatch/dispatch.impl.h
jstrzebonski: I completely agree, thanks for spotting this!

jstrzebonski: @PhilMiller done:

```
Unpacking wrong type, got=int (idx=3724543661), expected= (idx=1231123)
#0 int
#1 checkpoint::tests::unit::AA
#2 checkpoint::tests::unit::BB
#3 checkpoint::tests::unit::CC
#4 checkpoint::tests::unit::DD
#5 checkpoint::tests::unit::EE
```

docs/md/checkpoint_learn_build.md
```diff
@@ -69,7 +75,7 @@ With these set, invoke the script with two arguments: the path to the
 $ checkpoint/ci/build_cpp.sh <full-path-to-checkpoint-source> <full-path-to-build-dir>
 ```
 
-\subsection docker-build Building with `docker` containerization
+\subsection docker-build Building with docker containerization
```
jstrzebonski, line 80: It fixes this:
image

jstrzebonski:
`docker-compose` constantly ends with error 137 on `nvcc-10.1, ubuntu, mpich` pipeline. That error indicates that the container run out of memory. @lifflander is there anything we can do about it?

Obviously, locally on my machine checkpoint compiles just fine in docker.

PhilMiller (approved):

PhilMiller:
This all looks good to me, and the nvcc 10.1 failure seems to have been spurious, so I'm re-running it.

I'll ask that Jonathan and others get a chance at a final review pass before hitting 'Merge'

PhilMiller:
Hrm, nope, I think compiling the Kokkos test with error checking is running nvcc out of memory

.github/workflows/dockerimage-nvcc-10-ubuntu-mpich.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       CHECKPOINT_WARNINGS_AS_ERRORS: 0
       CHECKPOINT_DOCS: 0
       CHECKPOINT_ASAN: 0
+      CHECKPOINT_SERIALIZATION_ERROR_CHECKING: 1
```
PhilMiller: Perhaps we should turn this one off.

jstrzebonski: That's exactly what I wanted to try.

lifflander (approved):
This all looks good to me!

tests/unit/test_serialiation_error_checking.cc
```diff
@@ -0,0 +1,160 @@
+/*
+//@HEADER
+// *****************************************************************************
+//
+//                    test_serialiation_error_checking.cc
```
PhilMiller: I just noticed, there's a typo in the name of this file, and hence in the header

PhilMiller (changes requested):

tests/unit/test_serialiation_error_checking.cc
jstrzebonski: Renamed.

PhilMiller (approved):

PR #202, opened 2021-04-23
197 Expediently handle serialization of Kokkos::View inacessible from host space

### LINKED ISSUE #197: Serialization for Kokkos containers residing in host-inaccessible memory space, expediently
Take the simple steps towards #196 of working with the mirror view.

PhilMiller:
Allocate and operate on a full-sized host-accessible View as needed. The mirroring and `deep_copy` calls are no-ops if the view is already host accessible.

Depends on Kokkos 3.6 plus https://github.com/kokkos/kokkos/pull/4805, or Kokkos `develop`

Fixes #197

### DIFF
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -228,7 +228,7 @@ inline void serialize(
   // accessible from HostSpace, so no mirroring and copying is
   // necessary right now
 #if 0
-  auto host_view = Kokkos::create_mirror_view(view);
+  auto host_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, view);
   if (s.isPacking()) {
     // Create and use an execution space to avoid a global Kokkos::fence()
     auto exec_space = Kokkos::HostSpace::execution_space{};
@@ -347,7 +347,7 @@ inline void serialize_impl(SerializerT& s, Kokkos::DynRankView<T,Args...>& view)
   s | init;
 
   if (init) {
-    auto host_view = Kokkos::create_mirror_view(view);
+    auto host_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, view);
     if (s.isPacking()) {
       // Create and use an execution space to avoid a global Kokkos::fence()
       auto exec_space = Kokkos::HostSpace::execution_space{};
@@ -484,7 +484,7 @@ inline void serialize_impl(SerializerT& s, Kokkos::View<T,Args...>& view) {
   s | init;
 
   if (init) {
-    auto host_view = Kokkos::create_mirror_view(view);
+    auto host_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, view);
     if (s.isPacking()) {
       // Create and use an execution space to avoid a global Kokkos::fence()
       auto exec_space = Kokkos::HostSpace::execution_space{};
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -224,9 +224,7 @@ inline void serialize(
 
   DEBUG_PRINT_CHECKPOINT(s, "label=%s: size=%zu\n", label.c_str(), view.size());
 
-  // The implementation of Kokkos DynamicView demands that it be
-  // accessible from HostSpace, so no mirroring and copying is
-  // necessary right now
+  // Kokkos::deep_copy between DynamicView instances is not yet implemented
 #if 0
   auto host_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, view);
   if (s.isPacking()) {
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -447,7 +447,10 @@ inline void serialize_impl(SerializerT& s, Kokkos::View<T,Args...>& view) {
   if (init) {
     auto host_view = Kokkos::create_mirror_view(view);
     if (s.isPacking()) {
-      Kokkos::deep_copy(host_view, view);
+      // Create and use an execution space to avoid a global Kokkos::fence()
+      auto exec_space = Kokkos::HostSpace{};
+      Kokkos::deep_copy(exec_space, host_view, view);
+      exec_space.fence();
     }
 
     // Serialize the actual data owned by the Kokkos::View
@@ -473,7 +476,10 @@ inline void serialize_impl(SerializerT& s, Kokkos::View<T,Args...>& view) {
     }
 
     if (s.isUnpacking()) {
-      Kokkos::deep_copy(view, host_view);
+      // Create and use an execution space to avoid a global Kokkos::fence()
+      auto exec_space = Kokkos::HostSpace{};
+      Kokkos::deep_copy(exec_space, view, host_view);
+      exec_space.fence();
     }
   }
 }
--- a/src/checkpoint/container/view_serialize.h
+++ b/src/checkpoint/container/view_serialize.h
@@ -227,7 +227,7 @@ inline void serialize(
   auto host_view = Kokkos::create_mirror_view(view);
   if (s.isPacking()) {
     // Create and use an execution space to avoid a global Kokkos::fence()
-    auto exec_space = Kokkos::HostSpace{};
+    auto exec_space = Kokkos::HostSpace::execution_space{};
     Kokkos::deep_copy(exec_space, host_view, view);
     exec_space.fence();
   }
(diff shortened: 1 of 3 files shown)


PhilMiller:
~~Still need to handle `DynamicView`, `DynRankView`, and the like.~~ Done

PhilMiller:
This'll require a patch for Kokkos, for `deep_copy` of `DynamicView` and `DynRankView` to take execution space parameters.

PhilMiller:
> This'll require a patch for Kokkos, for `deep_copy` of `DynamicView` and `DynRankView` to take execution space parameters.

WIP in https://github.com/kokkos/kokkos/issues/4057

PhilMiller:
It turns out that `DyanmicView` is currently defined as always accessible on the host side
```c++
/** \brief  Must be accessible everywhere */
  using HostMirror = DynamicView;
```
So no copy dance is presently needed. If/when that changes, it will need to be addressed

PhilMiller:
> > This'll require a patch for Kokkos, for `deep_copy` of `DynamicView` and `DynRankView` to take execution space parameters.
> 
> WIP in [kokkos/kokkos#4057](https://github.com/kokkos/kokkos/issues/4057)

I think that's essentially done. Once it's merged and synced its way downstream, we can go ahead with this, which I think is now ready.

PhilMiller:
An ideal way to test this, independent of actual GPU hardware or whatever, is through facilities that I think Kokkos provides, or simulating memory spaces that are only accessible from kernel or non-kernel execution

PhilMiller:
> An ideal way to test this, independent of actual GPU hardware or whatever, is through facilities that I think Kokkos provides, or simulating memory spaces that are only accessible from kernel or non-kernel execution

It's here: https://github.com/kokkos/kokkos/pull/3546 -  core/src/Kokkos_LogicalSpaces.hpp

lifflander:
> > An ideal way to test this, independent of actual GPU hardware or whatever, is through facilities that I think Kokkos provides, or simulating memory spaces that are only accessible from kernel or non-kernel execution
> 
> It's here: [kokkos/kokkos#3546](https://github.com/kokkos/kokkos/pull/3546) - core/src/Kokkos_LogicalSpaces.hpp

@cz4rs Can you work on writing this test that shows it failing?

PhilMiller:
The `deep_copy` overloads taking execution spaces are integrated, and being released in Kokkos 3.6. That'll be in Trilinos and synced downstream by the time this is ready to make its way into use as well.

src/checkpoint/container/view_serialize.h
```diff
@@ -324,25 +348,33 @@ inline void serialize_impl(SerializerT& s, Kokkos::DynRankView<T,Args...>& view)
   s | init;
 
   if (init) {
+    auto host_view = Kokkos::create_mirror_view(view);
```
PhilMiller: We should pass `WithoutInitializing` and maybe an exec space for this

PhilMiller:
Syncing this into EMPIRE will want https://github.com/kokkos/kokkos/pull/4805 cherry-picked into its mirror of Trilinos, after the Kokkos 3.6 release lands in Trilinos.

src/checkpoint/container/view_serialize.h
PhilMiller: In general, `WithoutInitializing` avoids kernel invocation. Allocation inside Kokkos still potentially synchronizes stuff, but easing that is much further out on the horizon.

src/checkpoint/container/view_serialize.h
```diff
@@ -224,6 +224,19 @@ inline void serialize(
 
   DEBUG_PRINT_CHECKPOINT(s, "label=%s: size=%zu\n", label.c_str(), view.size());
 
+  // Kokkos::deep_copy between DynamicView instances is not yet implemented
```
PhilMiller, line 245: Follow-on PR can import the `DualView` mirror-and-sync approach taken in EMPIRE

src/checkpoint/container/view_serialize.h
```diff
@@ -475,7 +521,7 @@ inline void serialize_const(SerializerT& s, Kokkos::View<T,Args...>& view) {
   using T_non_const = typename ViewType::traits::non_const_data_type;
   Kokkos::View<T_non_const,Args...> tmp_non_const(view.label(), view.layout());
   if (s.isPacking() || s.isSizing()) {
```
PhilMiller: @lifflander As an aside to this PR, why does this call the copy for the sizing case?

PhilMiller:
I managed to write up the beginnings of a test using `LogicalMemorySpace`, and found that the overall serialization flow can't handle the mirror view stuff quite right. It's explicitly passing template arguments describing the original view somewhere that don't match the mirror view type.

PhilMiller:
Ok, test now demonstrates what it's supposed to - we can serialize and deserialize the contents of a View whose memory space isn't host-accessible.

src/checkpoint/container/view_serialize.h
```diff
@@ -224,6 +236,16 @@ inline void serialize(
 
   DEBUG_PRINT_CHECKPOINT(s, "label=%s: size=%zu\n", label.c_str(), view.size());
 
+  // Kokkos::deep_copy between DynamicView instances is not yet implemented
+#if 0
```
rppawlo, line 246: Do you need to keep this block around since it is explicitly disabled?

src/checkpoint/container/view_serialize.h
```diff
@@ -238,9 +260,15 @@ inline void serialize(
...
     };
 
-    TraverseRecursive<ViewType,T,1,decltype(fn)>::apply(view,fn);
+    TraverseRecursive<ViewType,T,1,decltype(fn)>::apply(host_view, fn);
 #else
-    TraverseManual<SerializerT,ViewType,1>::apply(s,view);
+    TraverseManual<SerializerT,ViewType,1>::apply(s, host_view);
+#endif
+
+#if 0
```
rppawlo, line 274: same question here

rppawlo (approved):
Thanks @PhilMiller !

src/checkpoint/container/view_serialize.h
PhilMiller: Roughly, it's the implementation that I want to provide when the underlying bits are ready

PhilMiller:
Marked as draft until #238 and #240 are merged.

src/checkpoint/container/view_serialize.h
PhilMiller: I looked at the history, and this has been there from the earliest sketch. It still doesn't make sense to me.

src/checkpoint/container/view_serialize.h
```diff
@@ -475,7 +531,7 @@ inline void serialize_const(SerializerT& s, Kokkos::View<T,Args...>& view) {
   using T_non_const = typename ViewType::traits::non_const_data_type;
   Kokkos::View<T_non_const,Args...> tmp_non_const(view.label(), view.layout());
   if (s.isPacking() || s.isSizing()) {
-    Kokkos::deep_copy(tmp_non_const, view);
+    deepCopyWithLocalFence(tmp_non_const, view);
```
PhilMiller, line 534: It's possible that we'd double-copy a device-space view of `const T`. We should see about folding those cases to a single copy, if possible. Perhaps by just aliasing the original const view with a non-const one, and skipping this step.

PhilMiller:
I'll test manually with Kokkos and CUDA, and then mark ready

src/checkpoint/container/view_serialize.h
PhilMiller: https://github.com/DARMA-tasking/magistrate/issues/245

src/checkpoint/container/view_serialize.h
```diff
@@ -93,6 +99,18 @@
...
 namespace checkpoint {
 
+namespace {
+
+template <typename T, typename U>
+void deepCopyWithLocalFence(T& dst, U& src) {
+  // Create and use an execution space to avoid a global Kokkos::fence()
+  auto exec_space = Kokkos::HostSpace::execution_space{};
+  Kokkos::deep_copy(exec_space, dst, src);
+  exec_space.fence();
```
PhilMiller, line 109: This may be effectively `Kokkos::fence()`, depending on implementation details. I'll try to see if there's an alternative

src/checkpoint/container/view_serialize.h
lifflander: I think that the isSizing can be removed.

PhilMiller: Will remove `|| isSizing()`

src/checkpoint/container/view_serialize.h
PhilMiller: I'll dig into this in a follow-up PR

lifflander (approved):

PR #208, opened 2021-05-13
#203 track and check memory usage during serialization

### LINKED ISSUE #203: Check if sizing, packing and unpacking reports/uses the same amount of memory

jstrzebonski:
Fixes #203 

As discussed on the weekly - the check is always active, and it's not recursive (yet).

---

Memory usage validation is done in two steps:
1. `validatePackerBufferSize()` called within `packBuffer()` compares memory calculated by Sizer with memory actually used by Packer.
2. At the end of the serialization Packer serializes the memory that it used. During deserialization, Unpacker deserialized Packer's memory usage and compares it with the memory amount that it actually used.

If any of this steps fail `buffer_size_error` is thrown.


jstrzebonski:
I'm not sure if the exceptions msges are good enough. If that's not the case I might need some help to improve them.

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -143,6 +149,35 @@ TraverserT Traverse::with(T& target, Args&&... args) {
...
+  } else if (t.isPacking()) {
+    // Size of buffer is serialized in next step, so actual memory usage
+    // needs to be calculated in advance
+    auto const usedBufferSize =
+  #if !defined(SERIALIZATION_ERROR_CHECKING)
+      t.usedBufferSize() + sizeof(SerialSizeType);
+  #else
+      t.usedBufferSize() + sizeof(SerialSizeType) +
+      sizeof(typeregistry::DecodedIndex);
+  #endif
```
jstrzebonski, line 161: That's somehow ugly but I don't see a better way of doing it correctly.

jstrzebonski: Good thing is that if for any reason this calculation becomes invalid, most of the tests will fail.

jstrzebonski:
One thing to note is that nvcc-11 constantly fails. It seems that there might be a problem with resources on GitHub pipeline.

PhilMiller:
As noted by Codecov, the error cases are not covered by the tests. Could you write tests that intentionally serialize different payloads between sizing and packing, and between packing and unpacking, and assert that an exception of the right type is thrown?

tests/unit/test_sizer.cc
```diff
@@ -65,16 +65,27 @@ struct Test1 {
 TEST_F(TestSizer, test_sizer_1) {
```
lifflander, line 65: I don't think I see any tests that check for the failure case when the buffer size doesn't match. Could we construct that use case which would improve code coverage?

lifflander (changes requested):
Overall this looks good. I would like more testing to cover cases where the buffer is not properly sized and expect failure. Hopefully, that will get code coverage to complain less also. Additionally, I would like to add documentation about this check and the error that it throws and what it means.

jstrzebonski:
@PhilMiller @lifflander 
I added missing unit tests. It's ready to re-review.

PhilMiller (approved):

PhilMiller:
I think the error message phrasing is quite good enough. If we wanted to be a bit more explicit, it could say "Check for a path mismatch in the associated `serialize` routine", and I think that would pretty much cover as much as any documentation that Jonathan was looking for would say.

lifflander (approved):

lifflander:
@jstrzebonski I've decided that the documentation of the error message is sufficient and thus merged this. We are waiting in EMPIRE for this to go through.

PR #212, opened 2021-06-04
#209 check memory usage recursively

### LINKED ISSUE #209: Recursively validate memory usage during serialization
Continuation of #203, only done recursively for each member of a serialized type.

jstrzebonski:
Fixes #209


jstrzebonski:
In case of an error in nested class, exemplary message looks like this:
```
For type 'int' serialization used 4B, but deserialization used 5B
#1 checkpoint::examples::E
#2 std::vector<checkpoint::examples::E, std::allocator<checkpoint::examples::E> >
#3 checkpoint::examples::D
#4 checkpoint::examples::C
#5 checkpoint::examples::B
#6 checkpoint::examples::A
#7 checkpoint::examples::MyTest
```

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -106,6 +105,36 @@ TraverserT& withTypeIdx(TraverserT& t) {
   return t;
 }
 
+template <typename T, typename TraverserT>
+TraverserT& withMemUsed(TraverserT& t, SerialSizeType len) {
+  using CleanT = typename CleanType<SerialSizeType>::CleanT;
```
lifflander: I don't like having another `CleanT` below that shadows this. I think it might be confusing.

I'm also confused why this statement is cleaning the `SerialSizeType`, which is a very simple type that basically is `size_t` or something. What am I missing here?

lifflander (changes requested):
Outside of my one comment, this looks good

src/checkpoint/dispatch/dispatch.impl.h
jstrzebonski: No, that's completely redundant, you're right! Removed.

PhilMiller:
I can't quite tell - is this replacing the top-level buffer usage check? Or is that preserved by the call to `withMemUsed`? We want that check to happen even if we're not doing any heavier weight verification.

jstrzebonski:
> I can't quite tell - is this replacing the top-level buffer usage check? Or is that preserved by the call to `withMemUsed`? We want that check to happen even if we're not doing any heavier weight verification.

If `SERIALIZATION_ERROR_CHECKING` is not set, the check will be done only for "top-level-type" (so compare memory usage for serialization of a whole object only) and if it set, it's done for the whole type hierarchy (for each object "component" separately and recursively).

jstrzebonski:
It's the same as for type registration.

jstrzebonski:
Take a look at `test_sizer.cc` where each test is split by ifdef.

PhilMiller (approved):

lifflander (approved):
Looks good to me now

PR #214, opened 2021-06-18
#161 use reconstruction logic for std::vector

### LINKED ISSUE #161: Unpacking of containers default-constructs elements rather than using reconstruction logic
For `std::vector<T> v`, the current code calls `s | len; v.resize(len)`, which results in calling the default constructor `T::T()`. This may not be valid code at all, for some types. For other types, the default constructor could be needlessly expensive. This violates the usual stated contract serialization offers, that reconstruction should use a tagged constructor if one is defined.

jstrzebonski:
Fixes #161

### DIFF
--- a/src/checkpoint/container/list_serialize.h
+++ b/src/checkpoint/container/list_serialize.h
@@ -45,6 +45,7 @@
 #define INCLUDED_CHECKPOINT_CONTAINER_LIST_SERIALIZE_H
 
 #include "checkpoint/common.h"
+#include "checkpoint/dispatch/allocator.h"
 #include "checkpoint/serializers/serializers_headers.h"
 #include "checkpoint/container/container_serialize.h"
 
@@ -60,17 +61,15 @@ inline typename std::enable_if_t<
 > deserializeOrderedElems(
   Serializer& s, ContainerT& cont, typename ContainerT::size_type size
 ) {
-  for (typename ContainerT::size_type i = 0; i < size; i++) {
-    #pragma GCC diagnostic push
-#if !defined(__has_warning)
-    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#elif __has_warning("-Wmaybe-uninitialized")
-    #pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
-    ElmT elm;
-    s | elm;
-    cont.push_back(std::move(elm));
-    #pragma GCC diagnostic pop
+  using Alloc = dispatch::Allocator<ElmT>;
+  using Reconstructor =
+    dispatch::Reconstructor<typename dispatch::CleanType<ElmT>::CleanT>;
+
+  Alloc allocated;
+  auto* reconstructed = Reconstructor::construct(allocated.buf);
+  cont.resize(size, *reconstructed);
+  for (auto& val : cont) {
+    s | val;
   }
 }
 
--- a/src/checkpoint/container/list_serialize.h
+++ b/src/checkpoint/container/list_serialize.h
@@ -56,10 +56,10 @@ namespace checkpoint {
 
 template <typename Serializer, typename ContainerT, typename ElmT>
 inline typename std::enable_if_t<
-  not std::is_same<Serializer, checkpoint::Footprinter>::value,
-  void
-> deserializeOrderedElems(
-  Serializer& s, ContainerT& cont, typename ContainerT::size_type size
+  not std::is_same<Serializer, checkpoint::Footprinter>::value, void>
+deserializeOrderedElems(
+  Serializer& s, ContainerT& cont, typename ContainerT::size_type size,
+  typename ReconstructorTraits<ElmT>::template isCopyConstructible<ElmT>* = nullptr
 ) {
   using Alloc = dispatch::Allocator<ElmT>;
   using Reconstructor =
@@ -70,6 +70,25 @@ inline typename std::enable_if_t<
   cont.resize(size, *reconstructed);
   for (auto& val : cont) {
     s | val;
+  }
+}
+
+template <typename Serializer, typename ContainerT, typename ElmT>
+inline typename std::enable_if_t<
+  not std::is_same<Serializer, checkpoint::Footprinter>::value, void>
+deserializeOrderedElems(
+  Serializer& s, ContainerT& cont, typename ContainerT::size_type size,
+  typename ReconstructorTraits<ElmT>::template isNotCopyConstructible<ElmT>* = nullptr
+) {
+  using Alloc = dispatch::Allocator<ElmT>;
+  using Reconstructor =
+    dispatch::Reconstructor<typename dispatch::CleanType<ElmT>::CleanT>;
+
+  Alloc allocated;
+  for (typename ContainerT::size_type i = 0; i < size; ++i) {
+    auto* reconstructed = Reconstructor::construct(allocated.buf);
+    s | *reconstructed;
+    cont.emplace_back(std::move(*reconstructed));
   }
 }
 
--- a/src/checkpoint/container/list_serialize.h
+++ b/src/checkpoint/container/list_serialize.h
@@ -56,10 +56,11 @@ namespace checkpoint {
 
 template <typename Serializer, typename ContainerT, typename ElmT>
 inline typename std::enable_if_t<
-  not std::is_same<Serializer, checkpoint::Footprinter>::value, void>
+  not std::is_same<Serializer, checkpoint::Footprinter>::value, void
+>
 deserializeOrderedElems(
   Serializer& s, ContainerT& cont, typename ContainerT::size_type size,
-  typename ReconstructorTraits<ElmT>::template isCopyConstructible<ElmT>* = nullptr
+  isCopyConstructible<ElmT>* = nullptr
 ) {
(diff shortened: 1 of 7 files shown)


jstrzebonski:
In general, I took reconstruct code almost 1:1 to be able to pass reconstruct tag directly to emplace_back(). Without that I would have to do `emplace_back(std::move(reconstructed_obj))` for all reconstruct cases, including tagged ctor, but if that's not a problem I think I could call `dispatch::Standard::construct()` directly, something like this:
```c++
template <typename T, typename VectorAllocator>
void reconstructVectorData(
  SerialSizeType const vec_size, std::vector<T, VectorAllocator>& vec) {
  auto* buf = dispatch::Standard::allocate<T>();
  for (SerialSizeType i = 0; i < vec_size; ++i) {
    auto* t = dispatch::Standard::construct<T>(buf);
    vec.emplace_back(std::move(*t));
  }
}
```

jstrzebonski:
@lifflander Do you think that's (more or less) the direction we wanted to head?

PhilMiller:
I think `vec.resize(vec_size, SERIALIZE_RECONSTRUCT_TAG)` is a feasible and preferable form for the case where we have the tagged reconstruction, since it avoids the intermediate allocation and move, and potentially even allows complete elision of all moves/copies if it gets inlined.

lifflander:
> @lifflander Do you think that's (more or less) the direction we wanted to head?

Sorry I was slow to reply. I agree with @PhilMiller that doing the resize is probably preferable for optimization.

src/checkpoint/container/vector_serialize.h
```diff
@@ -54,29 +57,57 @@ namespace checkpoint {
...
+  SerialSizeType const vec_size, std::vector<T, VectorAllocator>& vec,
+  typename ReconstructorTraits<T>::template isNotDefaultConsType<T>* = nullptr
+) {
+  using Alloc = dispatch::Allocator<T> const;
+  using CleanT =
+    dispatch::Reconstructor<typename dispatch::CleanType<T>::CleanT>;
+
+  Alloc allocated;
+  auto* reconstructed = CleanT::construct(allocated.buf);
+  vec.resize(vec_size, *reconstructed);
```
jstrzebonski, line 100: If `T` is not default-constructible reconstruct `T` using `Reconstructor` and resize vector.

Note - it uses copy ctor here. Would it make sense to have another function for `T` that's not copy-constructible but is move-constructible, and reconstruct vector using for-loop and `emplace_back()` + `std::move()`?

src/checkpoint/container/vector_serialize.h
```diff
@@ -54,29 +57,57 @@ namespace checkpoint {
...
+  typename ReconstructorTraits<T>::template isNotDefaultConsType<T>* = nullptr
+) {
+  using Alloc = dispatch::Allocator<T> const;
+  using Reconstructor =
+    dispatch::Reconstructor<typename dispatch::CleanType<T>::CleanT>;
+
+  Alloc allocated;
+  auto* reconstructed = Reconstructor::construct(allocated.buf);
+  vec.resize(vec_size, *reconstructed);
 }
```
jstrzebonski, line 101: If `T` is not default-constructible reconstruct `T` using `Reconstructor` and resize vector.

Note - it uses copy ctor here. Would it make sense to have another function for `T` that's not copy-constructible but is move-constructible, and reconstruct vector using for-loop and `emplace_back()` + `std::move()`?

src/checkpoint/container/vector_serialize.h
```diff
@@ -53,29 +56,57 @@ namespace checkpoint {
...
+  typename ReconstructorTraits<T>::template isNotDefaultConsType<T>* = nullptr
+) {
+  using Alloc = dispatch::Allocator<T>;
+  using Reconstructor =
+    dispatch::Reconstructor<typename dispatch::CleanType<T>::CleanT>;
+
+  Alloc allocated;
+  auto* reconstructed = Reconstructor::construct(allocated.buf);
+  vec.resize(vec_size, *reconstructed);
 }
```
jstrzebonski: If `T` is not default-constructible reconstruct `T` using `Reconstructor` and resize vector.

Note - it uses copy ctor here. Would it make sense to have another function for `T` that's not copy-constructible but is move-constructible, and reconstruct vector using for-loop and `emplace_back()` + `std::move()`?

src/checkpoint/container/list_serialize.h
```diff
@@ -60,17 +61,15 @@ inline typename std::enable_if_t<
...
+  using Alloc = dispatch::Allocator<ElmT>;
+  using Reconstructor =
+    dispatch::Reconstructor<typename dispatch::CleanType<ElmT>::CleanT>;
+
+  Alloc allocated;
+  auto* reconstructed = Reconstructor::construct(allocated.buf);
+  cont.resize(size, *reconstructed);
+  for (auto& val : cont) {
+    s | val;
   }
```
jstrzebonski: Would it make sense to have an additional function that can handle `T` that's not copy-constructible?

src/checkpoint/container/vector_serialize.h
jstrzebonski: Added.

src/checkpoint/container/list_serialize.h
jstrzebonski: Added.

PhilMiller (approved):

src/checkpoint/container/vector_serialize.h
```diff
@@ -53,43 +56,80 @@ namespace checkpoint {
...
+  not std::is_same<SerializerT, checkpoint::Footprinter>::value, void
+>
+serialize(SerializerT& s, std::vector<bool, VectorAllocator>& vec) {
+  auto const vec_size = serializeVectorMeta(s, vec);
+
+  if (s.isUnpacking()) {
+    constructVectorData(vec_size, vec);
+  }
 
   if (!s.isUnpacking()) {
```
lifflander, line 134: It there a more efficient way to serialize the booleans than iterating through them even if we don't know the exact type?

lifflander (approved):
I think the approach makes sense. The only overload that might be improved by another issue is the vector<bool> case that seems very un-optimized.

src/checkpoint/container/vector_serialize.h
jstrzebonski: The problem with `std::vector<bool>` is that it is implementation-defined. There is no way of getting an underlying array (no `data()` function), and no way of telling how much memory it occupies.

From the top of my head, I don't know how to do it better. I need to think it through.

jstrzebonski: I've just opened an issue for that https://github.com/DARMA-tasking/checkpoint/issues/229

PhilMiller: I think it would be perfectly acceptable to overstate the footprint of a vector of bool, say by reporting v.capacity()*sizeof bool, especially if it avoids worse contortions. I think it would also be acceptable for it to use the size of whatever proxy type gets returned

jstrzebonski: Yeah, but this function is for **not** footprinting.

lifflander: Ok, I understand the limitation. We might just not be able to do anything more efficient in this case.

src/checkpoint/container/list_serialize.h
```diff
@@ -55,30 +56,49 @@ namespace checkpoint {
...
 
 template <typename Serializer, typename ContainerT, typename ElmT>
 inline typename std::enable_if_t<
-  std::is_same<Serializer, checkpoint::Footprinter>::value,
-  void
-> deserializeOrderedElems(
+  std::is_same<Serializer, checkpoint::Footprinter>::value, void
+>
+deserializeOrderedElems(
   Serializer& s, ContainerT& cont, typename ContainerT::size_type size
```
PhilMiller, line 102: A potential refactoring, based on comments I had pointed out to me elsewhere: this case can avoid using `enable_if`, and just be an overload taking `checkpoint::Footprinter` rather than templating over `Serializer` at all.

src/checkpoint/container/vector_serialize.h
```diff
@@ -104,6 +144,25 @@ void serialize(Serializer& s, std::vector<bool, VectorAllocator>& vec) {
...
+  dispatch::serializeArray(s, vec.data(), vec.size());
+  s.addBytes(sizeof(T) * (vec.capacity() - vec.size()));
+}
+
+template <typename SerializerT, typename VectorAllocator>
+typename std::enable_if_t<
+  std::is_same<SerializerT, checkpoint::Footprinter>::value, void
+>
+serialize(SerializerT& s, std::vector<bool, VectorAllocator>& vec) {
+  s.countBytes(vec);
```
PhilMiller, line 162: This could/should also `s.addBytes(vec.capacity()*sizeof(bool))` or `s.addBytes(vec.capacity()*sizeof(bool)/CHAR_BIT)`.

I'm unconcerned which exactly, but I'd like something whose resulting value is `O(N)` so that if we or an application have code with an overgrown `vector<bool>`, it shows up.

src/checkpoint/container/list_serialize.h
PhilMiller: The same applies for related cases in the other files, too.

PR #216, opened 2021-06-28
215: move checkpoint version number into a separate file

### LINKED ISSUE #215: move `checkpoint` version number into a separate file
Follow-up to https://github.com/DARMA-tasking/vt/issues/1467.

cz4rs:
fixes #215


LICENSE
```diff
@@ -1,5 +1,6 @@
                           DARMA Toolkit v. 1.0.0
```
JacobDomagala, line 1: Shouldn't this line be also removed?

CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.8)
-project(checkpoint)
+project(checkpoint VERSION 1.1.0)
```
nlslatt, line 2: If there are no merges with checkpoint during a given week, are we still going to bump the version to keep it in sync with vt? Given the various packages are currently falling under the 1.0.0 toolkit, I don't think it's critical that the checkpoint version match the vt version. Given that, is 1.1.0 the right place to start with checkpoint? I'm okay either way, but think we should probably put a little thought into this if we haven't already.

LICENSE
nlslatt: This line needs to stay for copyright purposes.

cz4rs: AFAIK this is required by copyright to be put at least in the license file (this is what Jonathan has used in Sandia copyright process).

CMakeLists.txt
cz4rs: There was a verbal agreement to keep it in sync with vt at the last DARMA weekly, but without much discussion.
No strong opinion here either.

Another thing to note:
- we currently test `vt` with `checkpoint`@`develop` (and with `detector`@`master`): 
https://github.com/DARMA-tasking/vt/blob/525c7d0ef5ad703629b55fdf2dcd91e2eab7956c/ci/build_cpp.sh#L10
- should `vt` releases point to specific versions of the libs?

cz4rs:
Intel builds fail because of expired license:
```
  The C compiler

    "/opt/intel/install/bin/icc"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /build/detector/build/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/usr/bin/ninja cmTC_e1d7f && [1/2] Building C object CMakeFiles/cmTC_e1d7f.dir/testCCompiler.c.o
    FAILED: CMakeFiles/cmTC_e1d7f.dir/testCCompiler.c.o 
    /opt/intel/install/bin/icc    -o CMakeFiles/cmTC_e1d7f.dir/testCCompiler.c.o -c testCCompiler.c
    
    Error: Product support for your (Comp-CL) license has expired.
    License file(s) used were (in this order):
    **  1.  /opt/intel/system_studio_2018/compilers_and_libraries_2018.4.253/linux/bin/intel64/../../Licenses
    **  2.  /root/Licenses
    **  3.  /opt/intel/licenses/COM____DDW8-CKHSP7GN.lic
    **  4.  /Users/Shared/Library/Application Support/Intel/Licenses
    **  5.  /opt/intel/system_studio_2018/compilers_and_libraries_2018.4.253/linux/bin/intel64/*.lic
    
    Please refer https://software.intel.com/en-us/faq/purchasing-renewing-upgrading#support-expiration  for more information..
    
    icc: error #10052: could not checkout FLEXlm license
    ninja: build stopped: subcommand failed.
```

lifflander (approved):

lifflander:
I pulled back strongly on the required checks so that Intel compilers won't block mergebility.

CMakeLists.txt
cz4rs: `checkpoint` version should be aligned with `vt` - confirmed on DARMA weekly.

PR #220, opened 2021-07-13
#218 fix serialization error checking's memory leak

### LINKED ISSUE #218: LeakSanitizer detects memory leaks in test_serialization_error_checking

jstrzebonski:
Fixes #218

### DIFF
--- a/src/checkpoint/dispatch/dispatch.impl.h
+++ b/src/checkpoint/dispatch/dispatch.impl.h
@@ -275,8 +275,11 @@ buffer::ImplReturnType serializeType(T& target, BufferObtainFnType fn) {
 template <typename T>
 T* deserializeType(SerialByteType* data, SerialByteType* allocBuf) {
   auto mem = allocBuf ? allocBuf : Standard::allocate<T>();
-  T* t_buf = Standard::construct<T>(mem);
-  return Standard::unpack<T, UnpackerBuffer<buffer::UserBuffer>>(t_buf, data);
+  auto t_buf = std::unique_ptr<T>(Standard::construct<T>(mem));
+  T* traverser =
+    Standard::unpack<T, UnpackerBuffer<buffer::UserBuffer>>(t_buf.get(), data);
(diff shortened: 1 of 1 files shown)


src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -275,8 +275,11 @@ buffer::ImplReturnType serializeType(T& target, BufferObtainFnType fn) {
 template <typename T>
 T* deserializeType(SerialByteType* data, SerialByteType* allocBuf) {
   auto mem = allocBuf ? allocBuf : Standard::allocate<T>();
-  T* t_buf = Standard::construct<T>(mem);
-  return Standard::unpack<T, UnpackerBuffer<buffer::UserBuffer>>(t_buf, data);
+  auto t_buf = std::unique_ptr<T>(Standard::construct<T>(mem));
+  T* traverser =
+    Standard::unpack<T, UnpackerBuffer<buffer::UserBuffer>>(t_buf.get(), data);
+  t_buf.release();
```
PhilMiller, line 281: The explicit `release()` call shouldn't be necessary, since it will deallocate when `t_buf` goes out of scope.

jstrzebonski: Yes, but that's what I don't want to happen.

The problem was (and maybe I should actually describe it) that if serialization error checking throws, `mem` allocated by `Standard::allocate()` wasn't deallocated, so I added a unique_ptr to protect us against that leak. If `Standard::unpack` finished successfully then unique_ptr is no longer necessary, that's why I release it.

Also `traverser` and `t_buf` point to the same memory - `unpack()` returns the buffer that it gets as a parameter.

PhilMiller: I see, looks good then.

PhilMiller (approved):

JacobDomagala (approved):

nlslatt (approved):

PR #222, opened 2021-07-19
#221 sanitizers errors should break unit tests

### LINKED ISSUE #221: Errors found by sanitizers don't break unit tests
Which makes that errors semi-invisible. Possible fix -> https://github.com/sfc-gh-ebrossard/googletest/blob/master/googletest/docs/advanced.md#sanitizer-integration

jstrzebonski:
Fixes #221

### DIFF
--- a/tests/unit/main.cc
+++ b/tests/unit/main.cc
@@ -65,6 +65,9 @@ void __asan_on_error() {
 }
 
 int main(int argc, char **argv) {
+  int k = 0x7fffffff;
+  k += argc;
+
   ::testing::InitGoogleTest(&argc, argv);
 #if HAVE_MPI
   ::testing::AddGlobalTestEnvironment(new MPIEnvironment(&argc, argv));
--- a/tests/unit/main.cc
+++ b/tests/unit/main.cc
@@ -65,9 +65,6 @@ void __asan_on_error() {
 }
 
 int main(int argc, char **argv) {
-  int k = 0x7fffffff;
-  k += argc;
-
   ::testing::InitGoogleTest(&argc, argv);
 #if HAVE_MPI
   ::testing::AddGlobalTestEnvironment(new MPIEnvironment(&argc, argv));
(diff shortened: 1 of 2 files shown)


tests/unit/test_footprinter.cc
```diff
@@ -177,15 +177,6 @@ TEST_F(TestFootprinter, test_fundamental_types) {
...
   }
-
-  {
-    int arr[] = {1, 2, 3, 4, 5};
-    EXPECT_EQ(
-      checkpoint::getMemoryFootprint(arr),
-      sizeof(arr)
-      // actually: sizeof(&arr) + sizeof(arr)
-    );
-  }
```
jstrzebonski, line 188: UBSAN found an error here:
```
reference binding to misaligned address 0x000200000001 for type 'int', which requires 4 byte alignment
```


`dispatch::with()` does `auto val = cleanType(&target);` which eventually does `std::decay` on C-array type. Simplifying - it casts it to a raw pointer, making serialization of an array not possible. That's why I'm removing this test.

tests/unit/main.cc
```diff
@@ -54,6 +54,16 @@
 #include <Kokkos_Core.hpp>
 #endif
 
+extern "C" {
+void __ubsan_on_report() {
+   FAIL() << "Encountered an undefined behavior sanitizer error";
+}
+
+void __asan_on_error() {
+  FAIL() << "Encountered an address sanitizer error";
```
cz4rs, line 63: can you post an example of how does such error look in the CI? (is the complete output from ASAN included?)

cz4rs (approved):
looks good to me!

JacobDomagala (approved):

tests/unit/main.cc
jstrzebonski: Yes. Take a look -> https://github.com/DARMA-tasking/checkpoint/pull/222/checks?check_run_id=3115530025#step:7:493

image

lifflander (approved):

PR #234, opened 2022-03-30
233: add missing `libunwind` dependency

### LINKED ISSUE #233: `PR tests with vt` build fails because of missing `libunwind`
`libunwind` was added as a dependency to `vt` and this is now causing `magistrate` CI to fail. This happens when `PR tests with vt` build is using `ubuntu-cpp-vt` image:
 
```
> Run docker-compose run ubuntu-cpp-vt
(...)
/vt/src/vt/configs/error/stack_out.cc:49:10: fatal error: 'libunwind.h' file not found
```

---
_Note: this does **not** mean that `libunwind` is a dependency for `magistrate` now._

cz4rs:
fixes #233


scripts/check_containers.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# Check if there were any dockerfiles modified in the latest pull request.
+
+diff_latest() {
+  git diff --name-only "$(git log -1 --merges --pretty=format:%H)"..HEAD
+}
+
+# Copy the diff output to stderr to list the modified files
+if diff_latest | tee /dev/stderr | grep -i dockerfile > /dev/null
```
PhilMiller: I think the `if ... | grep >/dev/null` can be written `if ... | grep -q`

PhilMiller (approved):

scripts/check_containers.sh
PhilMiller: Your choice whether to make that change or not.

cz4rs: Looks better with `-q` I think, added that :+1:

PR #242, opened 2022-06-01
241 Drop old compilers

### LINKED ISSUE #241: Drop Clang 3.9, GCC 6/7 from CI, as they won't support C++17

PhilMiller:
Fixes #241


lifflander (approved):

nlslatt:
@PhilMiller If we're dropping support, we need to update the compiler version checks to make it clear we don't support those anymore.

PR #246, opened 2022-06-02
245 Serialize Kokkos::View<const T> through a non-const alias, rather than copying

### LINKED ISSUE #245: Avoid double-copying in serialize(View<const T, DeviceSpace>)
We have full-View-size bounce buffer serialization to handle Views with each of `const` element type and whose contents live in host-inaccessible device memory space (e.g. `CudaSpace` sans UVM). It looks like the code as currently written may naively allocate a mirror and copy the whole View to unwrap each of these.

Given that we're potentially serializing `View<T const>` already, I don't think we need to worry about the underlying storage not actually being writable. So, I think the right solution to this would be to not copy `View<T const>`, but rather to just alias it by an equivalent `View<T>`

PhilMiller:
Fixes #245


PhilMiller:
This is probably tripping over kokkos/kokkos#4913 and issues with ownership of an aliased allocation backing a `View`. I can do better.

PR #251, opened 2022-06-28
230 implement Kokkos::UnorderedMap serializer

### LINKED ISSUE #230: Provide a serializer for Kokkos::unordered_map

thearusable:
Closes #230

### DIFF
--- a/src/checkpoint/checkpoint.h
+++ b/src/checkpoint/checkpoint.h
@@ -65,6 +65,8 @@
 #include "checkpoint/container/unique_ptr_serialize.h"
 #include "checkpoint/container/view_serialize.h"
 
+#include <checkpoint/container/kokkos_unordered_map_serialize.h>
+
 #include "checkpoint/checkpoint_api.h"
 #include "checkpoint/checkpoint_api.impl.h"
 
--- a/src/checkpoint/checkpoint.h
+++ b/src/checkpoint/checkpoint.h
@@ -65,7 +65,7 @@
 #include "checkpoint/container/unique_ptr_serialize.h"
 #include "checkpoint/container/view_serialize.h"
 
-#include <checkpoint/container/kokkos_unordered_map_serialize.h>
+#include "checkpoint/container/kokkos_unordered_map_serialize.h"
 
 #include "checkpoint/checkpoint_api.h"
 #include "checkpoint/checkpoint_api.impl.h"
--- a/src/checkpoint/container/kokkos_unordered_map_serialize.h
+++ b/src/checkpoint/container/kokkos_unordered_map_serialize.h
@@ -85,7 +85,9 @@ template <
 void deserializeInsertElems(
   Serializer& s, Kokkos::UnorderedMap<Key, Value, Device, Hasher, EqualTo>& map,
   typename Kokkos::UnorderedMap<Key, Value, Device, Hasher, EqualTo>::size_type
-    map_size
+    map_size,
+  typename Kokkos::UnorderedMap<Key, Value, Device, Hasher, EqualTo>::size_type
+    map_capacity
 ) {
   using KeyAlloc = dispatch::Allocator<Key>;
   using ValueAlloc = dispatch::Allocator<Value>;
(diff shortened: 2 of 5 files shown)


PhilMiller:
Without having looked into the container's implementation, is there some notion of capacity that we should preserve across serialization?

thearusable:
@PhilMiller There is an option to pass `capacity_hint` to constructor, so that could increase the performance of deserialization slightly. If we don't provide capacity hint then the default 128 will be used

PhilMiller:
If there's a way to query it serialization time, I'd recommend doing that and then using it as a capacity hint at reconstruction

src/checkpoint/checkpoint.h
```diff
@@ -65,6 +65,8 @@
 #include "checkpoint/container/unique_ptr_serialize.h"
 #include "checkpoint/container/view_serialize.h"
 
+#include <checkpoint/container/kokkos_unordered_map_serialize.h>
```
PhilMiller: Please match existing use of quotes
```suggestion
#include "checkpoint/container/kokkos_unordered_map_serialize.h"
```

src/checkpoint/container/kokkos_unordered_map_serialize.h
```diff
@@ -0,0 +1,159 @@
...
+  std::is_same<SerializerT, checkpoint::Footprinter>::value, void
+> serialize(
+  SerializerT& s,
+  Kokkos::UnorderedMap<Key, Value, Device, Hasher, EqualTo>& map
+) {
+  using UnorderedMapType =
+    Kokkos::UnorderedMap<Key, Value, Device, Hasher, EqualTo>;
+
+  typename UnorderedMapType::size_type size = serializeContainerSize(s, map);
+  s.addBytes(size * (sizeof(Key) + sizeof(Value)));
```
cz4rs: is it possible to take capacity into account to improve precision?
see `std::vector` for example:
https://github.com/DARMA-tasking/magistrate/blob/594584f5ab51c9d5610c6a608f5a94e3c81d7c7d/src/checkpoint/container/vector_serialize.h#L147-L155

thearusable: @cz4rs I made a modification in this part of code to produce more accurate estimate of the UnorderedMap footprint. 
It's hard to precisely estimate actual size of the UnorderedMap without depending too much on the implementation details.

tests/unit/test_kokkos_serialize_unordered_map.cc
```diff
@@ -0,0 +1,111 @@
...
+
+  ASSERT_FALSE(refMap.failed_insert());
+
+  auto serialized = checkpoint::serialize<UnorderedMapType>(refMap);
+  auto deserialized =
+    checkpoint::deserialize<UnorderedMapType>(serialized->getBuffer());
+  auto& outMap = *deserialized;
+
+  ASSERT_FALSE(outMap.failed_insert());
+  ASSERT_EQ(refMap.size(), outMap.size());
```
PhilMiller, line 70:
```suggestion
  ASSERT_EQ(refMap.size(), outMap.size());
  ASSERT_EQ(refMap.capacity(), outMap.capacity());
```

PhilMiller (approved):

tests/unit/test_kokkos_serialize_unordered_map.cc
```diff
@@ -0,0 +1,112 @@
...
+  ASSERT_EQ(refMap.capacity(), outMap.capacity());
```
PhilMiller: Looks like these need to be `ASSERT_LE` instead, since the capacity ends up being greater than requested.

src/checkpoint/container/kokkos_unordered_map_serialize.h
cz4rs: :+1:

cz4rs (approved):
Looks good to me (except for the failing assertion) :+1:

tests/unit/test_kokkos_serialize_unordered_map.cc
thearusable: @PhilMiller Fixed. I forgot that capacity_hint will be increased by 16% and then rounded to nears multiple of 128, that's why we ended up with higher capacity than expected.

PhilMiller (approved):

PR #267, opened 2022-10-24
#264: Add serializers for Kokkos::pair and Kokkos::complex

### LINKED ISSUE #264: Add serializers for additional Kokkos types
Add serializers for Kokkos types:
- Kokkos::pair
- Kokkos::complex

thearusable:
Fixes #264

### DIFF
--- a/src/checkpoint/container/kokkos_complex_serialize.h
+++ b/src/checkpoint/container/kokkos_complex_serialize.h
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                 checkpoint.h
+//                      kokkos_complex_serialize.h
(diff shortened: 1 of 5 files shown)


PhilMiller:
These should be simple enough that you can just do it in one PR, and say this PR fixes the issue

PhilMiller (changes requested):
The (de)serialization logic is way more complicated than I would have expected. Could you write a bit about why that's all necessary, and maybe add that motivation as comments in the code?

thearusable:
@PhilMiller Indeed! That could be accomplished in much simpler way. Thanks for catching that!

PhilMiller (approved):

PR #271, opened 2022-11-21
Deserialize Kokkos::Views into existing allocation where possible

Matthew-Whitlock:
When calling deserializeInPlace flavored APIs, if the view to be deserialized into already has the label and dimensions that it will need after deserializing, don't create a new view to copy data into.

At a base level, this is a minor optimization for the cases of pre-allocated/initialized views (skipping a malloc, and whatever Kokkos steps are involved in registering a new view). However, it also enables keeping shallow copies of the input view "intact" (ie they continue to be shallow copies as expected, rather than now pointing to an old view allocation).

This leads to slightly more complex semantics for deserializeInPlace, since the creation of a new view or not depends on the input view and the serialized information. However, in cases where a user does not know apriori the label/dimensions of a view it seems unlikely that they would have shallow copies scattered about already anyway. And depending on deserializeInPlace to generate a new view seems like an odd use-pattern that can be easily fixed by using a typical deserialize function and copying the new view reference. With this in mind, I don't expect this would be a breaking change for applications. 

The specific benefit to this is the ability to use Magistrate for recovering views from a copy of the view, which is used by Kokkos Resilience's ViewHooks strategy for automatically detecting views to checkpoint. Without this, a deserialize has to be followed by an unnecessary copy into the original view's allocation.

### DIFF
--- a/src/checkpoint/container/view_equality.h
+++ b/src/checkpoint/container/view_equality.h
@@ -280,8 +280,8 @@ struct ViewEquality {
    */
   template <typename EqT = detail::DefaultEQ>
   static bool compareExtents(ViewT const& v1, ViewU const& v2, EqT eq = {}) {
-    auto const ndims_v1 = CountDims<ViewT>::numDims(v1);
-    auto const ndims_v2 = CountDims<ViewU>::numDims(v2);
+    auto const ndims_v1 = CountDims<ViewT>::numDims;
+    auto const ndims_v2 = CountDims<ViewU>::numDims;
     CHECKPOINT_APPLY_OP(eq, ndims_v1, ndims_v2, "num dims equal");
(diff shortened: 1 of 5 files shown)


PhilMiller:
I'm ambivalent about whether we actually want to directly support maintaining aliases when serializing a View. I believe there's a separate routine in the interface for serializing View contents that would be more appropriate to call from higher level logic ensuring this.

I'm not really working this week. Let's plan to discuss this next week when most of the crew is back from holidays.

PhilMiller:
https://github.com/DARMA-tasking/magistrate/blob/beaf39c7ec757b163afb080e6d9f4de2eb0fbcab/src/checkpoint/container/view_serialize.h#L698-L704

Matthew-Whitlock:
This is the functionality I was hoping for, thanks!

PR #276, opened 2023-01-19
#275: reconstruct: add new reconstruction for tuple elements with specialization

### LINKED ISSUE #275: Define reconstruction for Pair and Tuple in terms of reconstruction of its elements
A pair/tuple should be as reconstructible as its members, either by default construction or tagged

lifflander:
Fixes #275

### DIFF
--- a/src/checkpoint/dispatch/reconstructor.h
+++ b/src/checkpoint/dispatch/reconstructor.h
@@ -51,7 +51,12 @@
 #include <tuple>
 #include <cstdlib>
 
-namespace checkpoint { namespace dispatch {
+namespace checkpoint {
+
+template <typename T>
+struct CheckpointReconstructor;
+
+namespace dispatch {
 
 template <typename T>
 struct Reconstructor {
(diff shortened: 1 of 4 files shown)


lifflander:
Hmm, we need C++17 enabled here

PhilMiller:
I think this is all good, but I want to look at it more closely before approving

src/checkpoint/dispatch/reconstructor.h
```diff
@@ -51,7 +51,12 @@
 #include <tuple>
 #include <cstdlib>
 
-namespace checkpoint { namespace dispatch {
+namespace checkpoint {
+
+template <typename T>
+struct CheckpointReconstructor;
```
PhilMiller: I don't think you need to repeat the declaration here - it's already included.

PhilMiller: Other than this, looks good to me.

PhilMiller:
nvcc 10 is definitely not going to work anymore, since it has no support for C++17. For nvcc 11, I think we may need a newer CMake version so that it gets passed `-std=c++17` and not `-std=gnu++1z`

src/checkpoint/dispatch/reconstructor.h
lifflander: Fixed

PhilMiller (approved):

PR #297, opened 2023-03-29
#296: Specify no-c++-extensions so that compiler command-line flags to nvcc avoid warnings

PhilMiller:
Fixes #296

PR #299, opened 2023-03-30
298 Drop manual `-fPIC` command line option applied across the board

### LINKED ISSUE #298: Respect CMake `BUILD_SHARED` property rather than blanket-adding `-fPIC`
`src/CMakeLists.txt:77:target_compile_options(${CHECKPOINT_LIBRARY} PUBLIC "-fPIC")`

That means that downstream code that's building statically will still pick up `-fPIC` from us, which may cause problems or performance impacts.

PhilMiller:
Fixes #298

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -74,7 +74,6 @@ add_library(
 
 add_library(${CHECKPOINT_LIBRARY_NS} ALIAS ${CHECKPOINT_LIBRARY})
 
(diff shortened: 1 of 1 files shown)


PhilMiller:
~~Draft mode until #297 is merged and this can be rebased accordingly.~~

cz4rs (approved):
Looks good to me.

cz4rs:
We could have one of the CI builds here enabling `BUILD_SHARED_LIBS`. Not a must have since this is also covered in vt's `clang-9` build.

PR #300, opened 2023-03-30
#254: Clean up reconstruction logic for View, DynRankView, and DynamicView

### LINKED ISSUE #254: Simplify View/DynRankView reconstruction after Kokkos 3.7 ships
Kokkos 3.7 will include changes to properly round-trip the layout associated with a View(https://github.com/kokkos/kokkos/pull/5188), and DRV will likely do the same https://github.com/kokkos/kokkos/pull/5026 Thus, we can simplify the logic in `constructRankedView` and its callees down the stack.

PhilMiller:
Fixes #254

### DIFF
--- a/ci/docker/ubuntu-18.04-clang-cpp.dockerfile
+++ b/ci/docker/ubuntu-18.04-clang-cpp.dockerfile
@@ -52,11 +52,11 @@ RUN ./gtest.sh 1.8.1 /pkgs
 ENV GTEST_ROOT=/pkgs/gtest/install
 
 COPY ./ci/deps/kokkos.sh kokkos.sh
-RUN ./kokkos.sh 3.6.00 /pkgs 0
+RUN ./kokkos.sh 4.0.00 /pkgs 0
 ENV KOKKOS_ROOT=/pkgs/kokkos/install/lib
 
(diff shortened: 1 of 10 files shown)


PhilMiller:
This depends on at least Kokkos 3.7. I don't want to charge ahead with merging yet without at least a little bit of thought about the test that I had to disable because of https://github.com/kokkos/kokkos/issues/5609

It may be worth just going ahead, and we can run the concrete sibling test on real CUDA systems. It was mostly there as a TDD thing.

PhilMiller:
On further consideration, I think we can leave that test disabled for now, and maybe turn it back on if/when Kokkos loosens up or otherwise improves usability of logical memory spaces.

cz4rs (approved):
Looks good to me!

cz4rs:
> On further consideration, I think we can leave that test disabled for now, and maybe turn it back on if/when Kokkos loosens up or otherwise improves usability of logical memory spaces.

Should we create a follow-up issue or maybe include a reference to https://github.com/kokkos/kokkos/issues/5609 in the disabled test?

PhilMiller:
Follow-up issue created. Merging.

PR #305, opened 2023-05-26
#302: Fix unwanted ccache export

### LINKED ISSUE #302: Don't export `ccache` usage to downstream projects
An application that embeds Magistrate and refers to it via `add_subdirectory` finds that it gets `ccache` picked up automatically, regardless of whether the person building the application wanted it, or whether that application has its own configuration mechanism to choose to use `ccache`. Magistrate should definitely not be doing this in the embedded case.

Offending CMake code is here:
https://github.com/DARMA-tasking/magistrate/blob/develop/CMakeLists.txt#L40

thearusable:
Fixes #302

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,10 +37,12 @@ if(checkpoint_serialization_error_checking_enabled)
   message(STATUS "Building with serialization error checking enabled")
 endif()
 
-# Try to find ccache to speed up compilation
-find_program(ccache_binary ccache)
-if (ccache_binary)
-  set_directory_properties(PROPERTIES RULE_LAUNCH_COMPILE "${ccache_binary}")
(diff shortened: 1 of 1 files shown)


nlslatt:
@thearusable What is the new behavior introduced by this PR? Does it avoid forcing ccache on the parent project but still use it for vt embedded in that other project's build?

thearusable:
@nlslatt Yes. After the latest changes it will respect the `CMAKE_CXX_COMPILER_LAUNCHER` parameter. It will search for ccache only when it's not defined. In other cases it will respect whatever was set in `CMAKE_CXX_COMPILER_LAUNCHER` for this or parent project.

thearusable:
After update `magistrate` will only search for `ccache` in standalone mode when `CMAKE_CXX_COMPILER_LAUNCHER` is not defined.

nlslatt (approved):
Looks great

PR #317, opened 2023-07-26
274 add serializer for c++17 std::variant

### LINKED ISSUE #274: add serializer for C++17 `std::variant`
See https://github.com/DARMA-tasking/vt/blob/develop/src/vt/utils/adt/union.h

lifflander:
Fixes #274

This adds a serializer for std::variant. I've also included changes to rename `Serializer` to `BaseSerializer` after a very annoying bug in my code due to a mistake in my template. 

I've made changes that allow building the docker images with ARM so I can run/test on my M2 mac.

### DIFF
--- a/ci/deps/cmake.sh
+++ b/ci/deps/cmake.sh
@@ -4,16 +4,33 @@ set -exo pipefail
 
 if test $# -lt 1
 then
-    echo "usage: ./$0 <cmake-version>"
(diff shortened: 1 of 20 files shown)


nlslatt (commented):
Mostly looks good but I think you may have missed some needed changes related to the base serializer rename.

PR #343, opened 2024-05-28
#272: Add unit test to cover empty base optimization when using tuple

### LINKED ISSUE #272: Unit test for `std::tuple` serialization
There can be a pitfall with `std::tuple` where the offset of each element is implementation defined rather than a contiguous buffer. For example, an empty struct tuple element might have a size of 0 instead of a size of 1 due to empty base class optimization. We should unit test this proactively in order to avoid issues before they come up

thearusable:
Closes #272

### DIFF
--- a/tests/unit/test_tuple.cc
+++ b/tests/unit/test_tuple.cc
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                              test_tuple.cc
+//                                test_tuple.cc
 //                 DARMA/checkpoint => Serialization Library
 //
 // Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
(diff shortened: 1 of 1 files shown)


nmm0 (commented):
Thanks for doing this. Any chance you could add a test with Base?

The reason I'm asking as that at least in gcc, tuple uses inheritance and therefore EBO:

https://godbolt.org/z/G1qjsEGcx

i.e. `sizeof(std::tuple<empty, int>) == 4` but `sizeof(empty)+sizeof(int) == 5`. So the serialized size might not be the size of the tuple

thearusable:
> Thanks for doing this. Any chance you could add a test with Base?
> 
> The reason I'm asking as that at least in gcc, tuple uses inheritance and therefore EBO:
> 
> https://godbolt.org/z/G1qjsEGcx
> 
> i.e. `sizeof(std::tuple<empty, int>) == 4` but `sizeof(empty)+sizeof(int) == 5`. So the serialized size might not be the size of the tuple

@nmm0 Sure I'll add this

thearusable:
@nmm0 I extended the tuple tests. I added checks for serialized buffer sizes and a test with Base type. 
Is that what you had in mind?

nmm0 (approved):
Looks good! Thanks for doing this

PR #345, opened 2024-05-30
344 implement user traits

### LINKED ISSUE #344: Implement User Traits
### UserTraits
**What**: Serialize/deserialize calls are now also templated on `UserTraits...`, which is passed on to the base serializer. Some helpers for adding/removing and checking for the existence of trait are added as well.
**Why**: User traits allow arbitrary customization of serialization behavior based on user-defined terms. This enables more flexibility in checkpointing than an extra boolean variable, and is more clear in user-space code than a pre-defined variable name (EG: `VT::RestoreProxies` instead of `s.isCheckpointRecovery()`)

### Serialization overriding
**What**: When a non-intrusive serializer and an intrusive serializer are both found, call the non-intrusive serializer.
**Why**: Non-intrusive serializers can follow up with calls to the intrusive copy, meaning non-intrusive serializers can be used as a form of serialization subscriber. Particularly with user traits, we can define non-intrusive serializers which only run when a particular trait is used on a given object, allowing third parties (E.G. KokkosResilience) to perform custom logic before serialization of arbitrary objects. This can also be used to generally attach to serialization of objects for the purposes of tracing, etc.

For examples of these changes in action, see checkpoint_example_user_traits.cc/hpp

Matthew-Whitlock:
Closes #344

### DIFF
--- a/examples/checkpoint_example_user_traits.cc
+++ b/examples/checkpoint_example_user_traits.cc
@@ -40,6 +40,9 @@
 // *****************************************************************************
 //@HEADER
 */
+
+/// [Serialization with user traits]
+
 #include "checkpoint/checkpoint.h"
 
 #include "checkpoint_example_user_traits.h"
@@ -59,3 +62,5 @@ int main(int, char**){
   auto s_info_h = checkpoint::serialize<TestObj, misc::namespace_trait>(obj);
   auto s_info_i = checkpoint::serialize<TestObj, misc::hook_all_trait>(obj);
 }
+
+/// [Serialization with user traits]
--- a/examples/checkpoint_example_user_traits.cc
+++ b/examples/checkpoint_example_user_traits.cc
@@ -42,7 +42,7 @@
 */
 #include "checkpoint/checkpoint.h"
 
-#include "checkpoint_example_user_traits.hpp"
+#include "checkpoint_example_user_traits.h"
 
 int main(int, char**){
   using namespace test;
--- a/examples/checkpoint_example_user_traits.cc
+++ b/examples/checkpoint_example_user_traits.cc
@@ -2,7 +2,7 @@
 //@HEADER
 // *****************************************************************************
 //
-//                                footprinter.h
+//                      checkpoint_example_user_traits.cc
 //                 DARMA/magistrate => Serialization Library
 //
 // Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
@@ -40,55 +40,22 @@
 // *****************************************************************************
 //@HEADER
 */
+#include "checkpoint/checkpoint.h"
 
-#if !defined INCLUDED_SRC_CHECKPOINT_SERIALIZERS_FOOTPRINTER_H
-#define INCLUDED_SRC_CHECKPOINT_SERIALIZERS_FOOTPRINTER_H
+#include "checkpoint_example_user_traits.hpp"
 
-#include "checkpoint/common.h"
-#include "checkpoint/serializers/base_serializer.h"
+int main(int, char**){
+  using namespace test;
+  TestObj obj;
 
-namespace checkpoint {
-
-struct Footprinter : BaseSerializer {
-  Footprinter() : BaseSerializer(ModeType::Footprinting) { }
-
-  SerialSizeType getMemoryFootprint() const {
-    return num_bytes_;
-  }
-  void contiguousBytes(void*, SerialSizeType size, SerialSizeType num_elms) {
-    num_bytes_ += size * num_elms;
-  }
-
-  template<typename T>
-  void countBytes(const T& t) {
-    num_bytes_ += sizeof(t);
-  }
-
-  void addBytes(std::size_t s) {
-    num_bytes_ += s;
-  }
-
-private:
-  SerialSizeType num_bytes_ = 0;
-};
-
-namespace {
-    template <typename>
-    struct is_footprinter_impl : public std::false_type {};
-
-    template <typename UserTraits>
-    struct is_footprinter_impl<SerializerRef<Footprinter, UserTraits>> : std::true_type {};
-
-    template <>
-    struct is_footprinter_impl<Footprinter> : public std::true_type {};
+  //Each invocation will be handled based on the traits attached
+  auto s_info_a = checkpoint::serialize(obj);
+  auto s_info_b = checkpoint::serialize<TestObj, checkpoint_trait>(obj);
+  auto s_info_c = checkpoint::serialize<TestObj, checkpoint_trait, checkpoint_trait>(obj);
+  auto s_info_d = checkpoint::serialize<TestObj, random_trait, checkpoint_trait>(obj);
+  auto s_info_e = checkpoint::serialize<TestObj, checkpoint_trait, random_trait>(obj);
+  auto s_info_f = checkpoint::serialize<TestObj, random_trait, random_trait>(obj);
+  auto s_info_g = checkpoint::serialize<TestObj, shallow_trait>(obj);
+  auto s_info_h = checkpoint::serialize<TestObj, misc::namespace_trait>(obj);
+  auto s_info_i = checkpoint::serialize<TestObj, misc::hook_all_trait>(obj);
 }
-
-template<typename U>
-using is_footprinter = is_footprinter_impl<std::decay_t<U>>;
-
-template<typename U>
-static constexpr const bool is_footprinter_v = is_footprinter<U>::value;
-
-} /* end namespace checkpoint */
-
-#endif /*INCLUDED_SRC_CHECKPOINT_SERIALIZERS_FOOTPRINTER_H*/
--- a/examples/checkpoint_example_user_traits.hpp
+++ b/examples/checkpoint_example_user_traits.hpp
@@ -1,0 +1,109 @@
+#include "checkpoint/checkpoint.h"
+
+struct checkpoint_trait {} CheckpointTrait;
+struct shallow_trait {} ShallowTrait;
+
+using checkpoint::has_user_traits_v;
+using checkpoint::has_any_user_traits_v;
+
+namespace test {
+  struct random_trait {} RandomTrait;
+
+  struct TestObj {
+    int a = 1;
+
+    TestObj() {}
+
+    template<
+      typename SerT,
+      typename std::enable_if_t<
+        not has_user_traits_v<SerT, shallow_trait>, void*
+      > = nullptr
+    >
+    void serialize(SerT& s){
+      if constexpr(has_user_traits_v<SerT, checkpoint_trait>){
+        if(s.isSizing()) printf("Customizing serialization for checkpoint\n");
+        s | a;
+      } else {
+        if(s.isSizing()) printf("Default serializing testObj\n");
+      }
+
+      static_assert(not has_user_traits_v<SerT, shallow_trait>,
+          "ShallowTrait should have been removed!\n");
+    }
+  };
+}
+
+namespace test {
+  template<
+    typename SerT,
+    typename std::enable_if_t<
+      has_user_traits_v<SerT, random_trait>, void*
+    > = nullptr
+  >
+  void serialize(SerT& s, TestObj& myObj){
+    if(s.isSizing()){
+      printf("Inserting random extra object serialization step! ");
+    }
+    myObj.serialize(s);
+  }
+
+  template<
+    typename SerT,
+    typename std::enable_if_t<
+      has_user_traits_v<SerT, shallow_trait>, void*
+    > = nullptr
+  >
+  void serialize(SerT& s, TestObj& myObj){
+    if(s.isSizing()) printf("Removing shallow trait before passing along!\n");
+    auto newS = s.template withoutTraits<shallow_trait>();
+    myObj.serialize(newS);
+  }
+}
+
+namespace misc {
+  template<
+    typename SerT,
+    typename std::enable_if_t<
+      has_user_traits_v<SerT, test::random_trait>, void*
+    > = nullptr
+  >
+  void serialize(SerT& s, test::TestObj& myObj){
+    if(s.isSizing()){
+      printf("Serializers in other namespaces don't usually get found ");
+    }
+    myObj.serialize(s);
+  }
+
+
+  const struct namespace_trait {} NamespaceTrait;
+  template<
+    typename SerT,
+    typename std::enable_if_t<
+      has_user_traits_v<SerT, namespace_trait>, void*
+    > = nullptr
+  >
+  void serialize(SerT& s, test::TestObj& myObj){
+    if(s.isSizing()){
+      printf("A misc:: trait means we can serialize from misc:: too: ");
+    }
+    myObj.serialize(s);
+  }
+
+
+  const struct hook_all_trait {} HookAllTrait;
+  template<
+    typename SerT,
+    typename T,
+    typename std::enable_if_t<
+      has_user_traits_v<SerT, hook_all_trait>, void*
+    > = nullptr
+  >
+  void serialize(SerT& s, T& myObj){
+    if(s.isSizing()){
+      printf("We can even add on a generic pre-serialize hook: ");
+    }
+    auto newS = s.template withoutTraits<hook_all_trait>();
+    myObj.serialize(newS);
+  }
+}
--- a/src/checkpoint/checkpoint_api.h
+++ b/src/checkpoint/checkpoint_api.h
@@ -78,7 +78,7 @@ using SerializedReturnType = std::unique_ptr<SerializedInfo>;
  * \return a \c std::unique_ptr to a \c SerializedInfo containing the buffer
  * with serialized data and the size of the buffer
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 SerializedReturnType serialize(T& target, BufferCallbackType fn = nullptr);
 
 /**
@@ -101,7 +101,7 @@ SerializedReturnType serialize(T& target, BufferCallbackType fn = nullptr);
  *
  * \return a pointer to the newly reified \c T based on bytes in \c buf
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 T* deserialize(char* buf, char* object_buf);
 
 /**
@@ -118,7 +118,7 @@ T* deserialize(char* buf, char* object_buf);
  *
  * \return a unique pointer to the newly reified \c T based on bytes in \c buf
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 std::unique_ptr<T> deserialize(char* buf);
 
 /**
@@ -132,7 +132,7 @@ std::unique_ptr<T> deserialize(char* buf);
  * \param[in] t a valid pointer to a \c T that has been user-allocated and
  * constructed
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 void deserializeInPlace(char* buf, T* t);
 
 /**
@@ -143,7 +143,7 @@ void deserializeInPlace(char* buf, T* t);
  *
  * \return a unique pointer to \c T that must be deallocated
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 std::unique_ptr<T> deserialize(SerializedReturnType&& in);
 
 /**
@@ -153,7 +153,7 @@ std::unique_ptr<T> deserialize(SerializedReturnType&& in);
  *
  * \return number of bytes for the \c target
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 std::size_t getSize(T& target);
 
 /**
@@ -170,7 +170,7 @@ std::size_t getSize(T& target);
  *
  * \return memory footprint of the \c target
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 std::size_t getMemoryFootprint(T& target, std::size_t size_offset = 0);
 
 /**
@@ -184,7 +184,7 @@ std::size_t getMemoryFootprint(T& target, std::size_t size_offset = 0);
  * \param[in] target the \c T to serialize
  * \param[in] file name of the file to create
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 void serializeToFile(T& target, std::string const& file);
 
 /**
@@ -200,7 +200,7 @@ void serializeToFile(T& target, std::string const& file);
  *
  * \return unique pointer to the new object \c T
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 std::unique_ptr<T> deserializeFromFile(std::string const& file);
 
 /**
@@ -214,7 +214,7 @@ std::unique_ptr<T> deserializeFromFile(std::string const& file);
  * \param[in] file the filename to read with bytes for \c T
  * \param[in] t a valid, constructed \c T to deserialize into
  */
-template <typename T>
+template <typename T, typename... UserTraits>
 void deserializeInPlaceFromFile(std::string const& file, T* buf);
 
 /**
@@ -227,7 +227,7 @@ void deserializeInPlaceFromFile(std::string const& file, T* buf);
  * \param[in] target the \c T to serialize
  * \param[in] stream to serialize into, with tellp and write functions.
  */
-template <typename T, typename StreamT>
+template <typename T, typename... UserTraits, typename StreamT>
 void serializeToStream(T& target, StreamT& stream);
 
 /**
@@ -243,7 +243,7 @@ void serializeToStream(T& target, StreamT& stream);
  *
  * \return unique pointer to the new object \c T
  */
-template <typename T, typename StreamT>
+template <typename T, typename... UserTraits, typename StreamT>
 std::unique_ptr<T> deserializeFromStream(StreamT& stream);
 
 /**
@@ -257,7 +257,7 @@ std::unique_ptr<T> deserializeFromStream(StreamT& stream);
  * \param[in] stream the stream to read with bytes for \c T, with tellg and read functions
  * \param[in] t a valid, constructed \c T to deserialize into
  */
-template <typename T, typename StreamT>
+template <typename T, typename... UserTraits, typename StreamT>
 void deserializeInPlaceFromStream(StreamT& stream, T* buf);
 
 
--- a/src/checkpoint/dispatch/dispatch.h
+++ b/src/checkpoint/dispatch/dispatch.h
@@ -98,7 +98,7 @@ struct Traverse {
    *
    * \return the constructed traverser after traversal is complete
    */
-  template <typename T, typename TraverserT, typename... Args>
+  template <typename T, typename TraverserT, typename UserTraits = UserTraitHolder<>, typename... Args>
   static TraverserT with(T& target, Args&&... args);
 
   /**
@@ -131,7 +131,7 @@ struct Standard {
    *
    * \return the number of bytes
    */
-  template <typename T, typename SizerT, typename...Args>
+  template <typename T, typename SizerT, typename UserTraits, typename...Args>
   static SerialSizeType size(T& target, Args&&... args);
 
   /**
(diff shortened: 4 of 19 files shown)


nmm0:
@Matthew-Whitlock will this require some changes to VT?

cz4rs (changes requested):
@Matthew-Whitlock `PR tests with vt` build is currently failing in the configuration stage and reports success incorrectly. #352 fixes both these issue, please rebase once it is merged.

tests/unit/test_footprinter.cc
```diff
@@ -105,9 +105,8 @@ struct TestDerived2 : TestBase {
   template <
     typename SerializerT,
     typename = std::enable_if_t<
-      std::is_same<
-        SerializerT,
-        checkpoint::Footprinter
+      checkpoint::is_footprinter<
```
cz4rs: Can we also get `checkpoint::is_footprinter_v` helper?

examples/checkpoint_example_user_traits.hpp
```diff
@@ -0,0 +1,66 @@
+#include "checkpoint/checkpoint.h"
+
+struct checkpoint_trait {} CheckpointTrait;
+struct shallow_trait {} ShallowTrait;
+
+namespace test {
+  struct random_trait {} RandomTrait;
+
+  struct TestObj {
```
nmm0: We should add this as a unit test (it can still also be an example)

src/checkpoint/checkpoint_api.h
```diff
@@ -78,7 +78,7 @@ using SerializedReturnType = std::unique_ptr<SerializedInfo>;
  * \return a \c std::unique_ptr to a \c SerializedInfo containing the buffer
  * with serialized data and the size of the buffer
  */
-template <typename T>
+template <typename... UserTraits, typename T>
```
nmm0: It's a bit weird to me that the template arguments here are in a different order compared to deserialize because of the argument type deduction (while deserialize can't deduce).

@lifflander and @Matthew-Whitlock what do y'all think of potentially making them deducable with something like:

```cpp
template< typename... UserTraits >
struct user_traits {};  // we could even have things like has_trait here
```

But then would have to add a new overload:
```cpp
template <typename T, typename... UserTraits>
SerializedReturnType serialize(T& target, user_traits<UserTraits...>, BufferCallbackType fn = nullptr) 
```

src/checkpoint/dispatch/dispatch.h
```diff
@@ -188,21 +188,21 @@ struct Standard {
...
 };
 
-template <typename T>
+template <typename T, typename... UserTraits>
 buffer::ImplReturnType packBuffer(
   T& target, SerialSizeType size, BufferObtainFnType fn
 );
 
 template <typename Serializer, typename T>
 inline void serializeArray(Serializer& s, T* array, SerialSizeType const len);
```
nmm0, line 196: should this have UserTraits also?

src/checkpoint/dispatch/vrt/virtual_serialize.h
```diff
@@ -62,6 +62,15 @@ void virtualSerialize(T*& base, SerializerT& s) {
...
   using BaseT = ::checkpoint::dispatch::vrt::checkpoint_base_type_t<T>;
   auto serializer_idx = serializer_registry::makeObjIdx<BaseT, SerializerT>();
+
+  if constexpr (not std::is_same_v<SerializerT, typename SerializerT::TraitlessT>){
+    static bool warned_once = false;
+    if(!warned_once){
+      fprintf(stderr, "Warning: Magistrate cannot support UserTraits being visible to virtually serialized objects.\n");
+      warned_once = true;
+    }
+  }
```
nmm0: Can this be a static assert? Also we should use libfmt here instead of fprintf

src/checkpoint/serializers/reference.h
```diff
@@ -0,0 +1,178 @@
...
+
+namespace {
+  template<typename T, typename... U>
+  struct first_type_matches : public std::false_type {};
+  template<typename T, typename... U>
+  struct first_type_matches<T, T, U...> : public std::true_type {};
+}
+
+template<typename SerT, typename... UserTraits>
+struct SerializerRef
```
nmm0: I think the file should be named `serializer_ref.h` because otherwise it kind of gives the impression that it's a serializer for references

src/checkpoint/serializers/reference.h
```diff
@@ -0,0 +1,178 @@
...
+  template<typename... Args>
+  explicit SerializerRef(Args&&... args)
+  {
+    if constexpr(first_type_matches<std::shared_ptr<SerT>, std::decay_t<Args>...>::value and sizeof...(Args) == 1) {
+        //If only 1 arg which is a shared pointer to the base serializer, use that instead of initializing a new serializer.
+        impl = {std::forward<Args>(args)...};
+    } else {
+        impl = std::make_shared<SerT>(std::forward<Args>(args)...);
+    }
+  }
```
nmm0: I think this should be two overloads:
```cpp
explicit SerializerRef(std::shared_ptr<SerT> ptr);

template<typename... Args>
explicit SerializerRef(std::in_place_t, Args&&... args);
```

Using the `std::in_place` to avoid ambiguity.

nmm0 (changes requested):
About half reviewed

src/checkpoint/checkpoint_api.h
nmm0: Example usage:

```cpp
serialize(my_int, user_traits<shallow>{});
```

src/checkpoint/serializers/reference.h
```diff
@@ -0,0 +1,178 @@
...
+  void skip(Args&&... args){
+    impl->skip( std::forward<Args>(args)... );
+  }
+  SerialByteType* getBuffer() const { return impl->getBuffer(); }
+  SerialByteType* getSpotIncrement(const SerialSizeType inc) { return impl->getSpotIncrement(inc); }
+  bool isVirtualDisabled() const { return impl->isVirtualDisabled(); }
+  void setVirtualDisabled(bool val) { impl->setVirtualDisabled(val); }
+
+
+
```
nmm0:
```suggestion
```

src/checkpoint/serializers/reference.h
```diff
@@ -0,0 +1,178 @@
...
+  SerT* operator->(){ return impl.get(); }
+
+  template<typename... Traits>
+  operator SerializerRef<SerT, Traits...>() const {
+    return SerializerRef<SerT, Traits...>(impl);
+  }
+
```
nmm0:
```suggestion
```

src/checkpoint/serializers/reference.h
```diff
@@ -0,0 +1,178 @@
...
+  //Returns a new reference with traits set to the inputs, regardless of this reference's traits.
+  template<typename... NewTraits>
+  SerializerRef<SerT, NewTraits...> setTraits(UserTraitHolder<NewTraits...> = {}){
+    return *this;
+  }
+
+private:
+  std::shared_ptr<SerT> impl;
```
nmm0: I guess a design question here is do we actually need a shared pointer? Can this hold ownership with a unique_ptr or even just the object itself and we move it in?

nmm0 (changes requested):

src/checkpoint/serializers/reference.h
nmm0: After discussing it, we decided to make this a raw ptr that doesn't hold ownership (i.e. the caller is responsible for creating the serializer and it shouldn't go out of scope)

tests/unit/test_footprinter.cc
Matthew-Whitlock: Done - I missed a spot to swap to is_footprinter_v in the footprinter unit test, but that's handled by #370 now

src/checkpoint/checkpoint_api.h
Matthew-Whitlock: I swapped the template ordering, but left it without the type deduction for now. Mostly to defer questions about where to place the new argument in the various overloads

src/checkpoint/dispatch/dispatch.h
Matthew-Whitlock: It should not have. It still have UserTraits, but now for good reason

src/checkpoint/dispatch/dispatch.impl.h
```diff
@@ -50,6 +50,7 @@
 
 #include <stdexcept>
 #include <string>
+#include <optional>
```
PhilMiller: `std::optional` is documented as being added in C++17, while [Magistrate documentation](https://darma-tasking.github.io/checkpoint_docs/html/index.html) says it's a C++14 library.

Fortunately, it doesn't look like `std::optional` is actually *used* anywhere in here, so this inclusion should just be dropped.

Also, maybe that line of the documentation is out of date, in which case, please open an issue. Additionally, I see two more issues of documentation markup not being rendered properly, which may represent additional issues(s):
* `\checkpoint`
* `@m_class{m-note m-dim}`

cz4rs: C++14 thing is a bug, I opened #371 to track that.

:+1: for dropping an unused `#include`.

nmm0:
@Matthew-Whitlock  Can you rebase and mark as ready for review?

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+  explicit UserObjectA(int val)
+    : name('A'), val_a(val) {};
+  explicit UserObjectA(int val, char name_)
+    : name(name_), val_a(val) {};
+
+  template <typename S>
+  void serialize(S& s) {
+    std::cout << "A: serializing with type "
+      << abi::__cxa_demangle(typeid(s).name(), nullptr, nullptr, nullptr)
+      << std::endl;
```
nmm0: I think you need to use `std::cerr` for this to be visible? Either way this is not portable, just print the mangled type directly or remove this

src/checkpoint/serializers/serializer_ref.h
```diff
@@ -0,0 +1,171 @@
...
+  //Big block of helpers for conveniently checking traits in different contexts.
+  template<typename... Traits>
+  using has_traits = bool_enable_if<TraitHolder::template has<Traits...>::value>;
+  template<typename... Traits>
+  using has_any_traits = bool_enable_if<TraitHolder::template has_any<Traits...>::value>;
+
+  template<typename... Traits>
+  using has_not_traits = bool_enable_if<!(has_traits<Traits...>::value)>;
+  template<typename... Traits>
+  using has_not_any_traits = bool_enable_if<!(has_any_traits<Traits...>::value)>;
```
nmm0: I think the best thing would be to provide these as stdlib-traits-like-things, e.g. outside of the class:

```cpp
template<typename S, typename... Traits>
struct has_traits : typename S::TraitHolder::template has<Traits...> {};

template<typename S, typename... Traits>
constexpr bool has_traits_v = has_traits<S, Traits>::value;
```
```

File: src/checkpoint/dispatch/dispatch_serializer_nonbyte.h [hunk 14]
```diff
@@ -91,7 +91,7 @@ struct SerializerDispatchNonByte {
 
   template <typename U>
   using hasInSerialize =
-  typename std::enable_if<SerializableTraits<U,S>::has_serialize_instrusive, T>::type;
+  typename std::enable_if<SerializableTraits<U,S>::has_serialize_instrusive && !SerializableTraits<U,S>::has_serialize_noninstrusive, T>::type;
```
nmm0, line 94: Good catch :)

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,196 @@
...
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+
+#if !defined INCLUDED_SRC_CHECKPOINT_TRAITS_USER_TRAITS_H
+#define INCLUDED_SRC_CHECKPOINT_TRAITS_USER_TRAITS_H
+
+namespace {
```
nmm0:
```suggestion
namespace detail {
```

The reason for this is because this is in a header, the anonymous namespace will just have a compiler-defined name that is different for each translation unit that includes this file, which is probably not what you want.

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,196 @@
...
+  struct NoTrait;
+
+  template<typename Traits, typename T, typename... U>
+  struct without_helper {
```
nmm0: Here and elsewhere in the PR make sure the casing matches the rest of magistrate, i.e. `CamelCase` for struct names

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,196 @@
...
+  };
+  template<typename Traits, typename T>
+  struct without_helper<Traits, T> {
+    using type = typename Traits::template _without_trait<T>;
+  };
+}
+
+namespace checkpoint {
+namespace SerializerUserTraits {
+
```
nmm0, line 49: I think this isn't mentioned in the style guide but I guess use snake case for the namespaces so they are consistent

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,196 @@
...
+  struct is_same_template
+    : std::false_type {};
+  template<template <typename...> typename Templ, typename... U, typename... V>
+  struct is_same_template<Templ<U...>, Templ<V...>>
+    : std::true_type {};
+
+  template<typename U, typename V>
+  struct traits_match
+    : std::disjunction<
+#ifdef CHECKPOINT_INHERITED_USER_TRAITS
```
nmm0: Are these macros defined anywhere?

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,196 @@
...
+    static inline constexpr bool has_v = has<T...>::value;
+    template<typename... T>
+    static inline constexpr bool has_any_v = has_any<T...>::value;
+
+    template<typename T, typename... U>
+    using without = typename without_helper<BaseTraits, T, U...>::type;
+  };
+
+  //Empty specialization w/ shortcuts.
+  template<>
```
nmm0, line 139: Does this speed up compilation time? If not we can probably remove it because of added complexity

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,196 @@
...
+
+    template<typename, typename...>
+    friend struct TraitHolder;
+    template<typename...>
+    friend struct without_helper;
+  };
+
+  //Specialization to remove nested traits.
+  template<typename... Traits, typename... NestedTraits>
+  struct TraitHolder<TraitHolder<NestedTraits...>, Traits...>
```
nmm0: How do these pop up? This feels like it is patching out something weird in a template instantiation somewhere -- do we need to support in the API nested traits that can be collapsed out?

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+
+
+struct UserObjectB : public UserObjectA {
+  UserObjectB() = default;
+  explicit UserObjectB(int val)
+    : UserObjectA(0, 'B'), val_b(val) {};
+
+  template <typename S>
+  void serialize(S& s) {
+    std::cout << "B: serializing with type "
```
nmm0: Same here

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+TYPED_TEST_CASE_P(TestUserTraits);
+
+struct CheckpointTrait {};
+struct CheckpointTraitNonintrusive {};
+namespace CheckpointNamespace {
+struct CheckpointTraitNamespaced {};
+}
+
+struct ShallowTrait {};
+struct TraitPairA {};
```
nmm0, line 65: `TraitPairA` is a little confusing as they don't seem the be pairs?

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+    s | val_a;
+  }
+
+  char name;
+  int val_a;
+};
+
+template<
+  typename S,
+  typename = typename S::template has_traits<CheckpointTraitNonintrusive>::type
```
nmm0: use `enable_if` here, eventually will be replaced by `requires`

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+>
+void serialize(S& s, UserObjectA& obj){
+  s | obj.name;
+  obj.serialize(s);
+}
+
+namespace CheckpointNamespace {
+template<
+  typename S,
+  typename = typename S::template has_traits<CheckpointTraitNamespaced>::type
```
nmm0: same here

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+
+TEST(TestUserTraits, test_trait_removal) {
+  UserObjectB objB(u_val);
+  int old_b_val = objB.val_b, old_a_val = objB.val_a;
+  int new_b_val = old_b_val+1, new_a_val = old_a_val+1;
+
+  auto serB = checkpoint::serialize<UserObjectB, ShallowTrait>(objB);
+  objB.val_b = new_b_val;
+  objB.val_a = new_a_val;
+  checkpoint::deserializeInPlace<UserObjectB, ShallowTrait>(serB->getBuffer(), &objB);
```
nmm0, line 181: should this test also check name to ensure that the UserObjectA was called correctly

tests/unit/test_user_traits.cc
```diff
@@ -0,0 +1,204 @@
...
+
+TEST(TestUserTraits, test_trait_addition) {
+  UserObjectB objB(u_val);
+  int old_b_val = objB.val_b, old_a_val = objB.val_a;
+  int new_b_val = old_b_val+1, new_a_val = old_a_val+1;
+
+  auto serB = checkpoint::serialize<UserObjectB, TraitPairA>(objB);
+  objB.val_b = new_b_val;
+  objB.val_a = new_a_val;
+  checkpoint::deserializeInPlace<UserObjectB, TraitPairA>(serB->getBuffer(), &objB);
```
nmm0, line 208: same here

tests/unit/test_user_traits.cc
nmm0: also do we test if this gets called?

nmm0 (changes requested):
Whew, sorry it took so long I added a bunch comments

tests/unit/test_user_traits.cc
Matthew-Whitlock: `std::cout` prints to the test log file, but I think it's hidden unless the test fails (or you pass an extra flag). I'm dropping the demangle though

src/checkpoint/serializers/serializer_ref.h
Matthew-Whitlock: I like the idea of keeping them inside the class just for better context in user code, but it does help avoid the NVCC compiler weirdness so I agree

src/checkpoint/traits/user_traits.h
Matthew-Whitlock: Happy to make this change, but I think underscore casing makes sense for type manipulation and template helpers. The rest of the code is a bit inconsistent, though.

serializable_traits.h has:
```
template <typename U>
using specialized_reconstruct_t = decltype(
  checkpoint::CheckpointReconstructor<U>::reconstruct(
    std::declval<U*&>(),
    std::declval<void*>()
  )
);
using has_specialized_reconstruct =
  detection::is_detected<specialized_reconstruct_t, T>;

//But also this style
template <typename U, typename V>
using serializeThis_t =
  decltype(std::declval<U>().serializeThis(std::declval<V&>()));
using has_serializeThis = detection::is_detected<serializeThis_t, T, S>;
```

checkpoint_traits.h:
```
template <typename U>    
using contiguousBytes_t = decltype(std::declval<U>().contiguousBytes(    
  std::declval<void*>(), std::declval<SizeType>(), std::declval<SizeType>()    
));    
using has_contiguousBytes = detection::is_detected<contiguousBytes_t, T>;
```

reconstructor_traits.h:
```
template <typename T>
using isDefaultConsType =
  std::enable_if_t<std::is_default_constructible<T>::value>;
```

nmm0: Hah, I'll defer to @lifflander then for this one :P

src/checkpoint/traits/user_traits.h
Matthew-Whitlock: They aren't - I wasn't sure if it would be best to have these extra checks or not, so I figured I'd defer the question with these. Think I should drop the ifdefs or drop the checks?

src/checkpoint/traits/user_traits.h
Matthew-Whitlock: Having a NoTrait instead of an empty variadic helps simplify the main logic by allowing us to always assume at least one type is in the TraitHolders being passed around

src/checkpoint/traits/user_traits.h
Matthew-Whitlock: I don't believe this does pop up naturally, I just added it on during my first pass thinking it would come in handy. I'll go ahead and drop these

tests/unit/test_user_traits.cc
Matthew-Whitlock: The naming is based on the check in UserObjectA that expects to have either both or neither of TraitPairA and TraitPairB

tests/unit/test_user_traits.cc
Matthew-Whitlock: We do check if this is called, in test_serialize_extra. It confirms that UserObjectA does not serialize the name without a given trait, and does when that trait is included. And here we test with the three traits:
```
REGISTER_TYPED_TEST_CASE_P(TestUserTraits, test_serialize_extra);    
using TraitTypes = ::testing::Types<CheckpointTrait, CheckpointTraitNonintrusive, CheckpointNamespace::CheckpointTraitNamespaced>;    
INSTANTIATE_TYPED_TEST_CASE_P(checkpoint_trait, TestUserTraits, TraitTypes, );
```

Technically we aren't verifying that the specific function is called based on the trait, but that would require our has_traits to be broken and I think we test that well enough with the checks for ShallowTrait removal and TraitPairA/B matching

tests/unit/test_user_traits.cc
Matthew-Whitlock: We can, but I don't think that's checking for anything that wouldn't already be caught by test_serialize_extra so I'd rather keep the checks within the test's scope

src/checkpoint/traits/user_traits.h
Matthew-Whitlock: Actually, this was helpful in the VT resilience updates to make a new checkpoint::serialize call with the same traits as another serializer

tests/unit/test_user_traits.cc
Matthew-Whitlock: I added an extra check to ensure that our trait detection doesn't mix up two different traits, so as long as that passes and this passes we can be sure that we call the extra functionality and that the right one is called

tests/unit/test_user_traits.cc
Matthew-Whitlock: With the extra check for correct trait detection, we should be fine leaving the name checking to the test_serialize_extra test

tests/unit/test_user_traits.cc
cz4rs: Makes sense to me.

tests/unit/test_user_traits.cc
nmm0: I guess the functionality it would test would be if somehow other traits are not added by accident.

src/checkpoint/traits/user_traits.h
nmm0: Hmm I would then put that in a separate metafunction instead of auto-collapsing

src/checkpoint/traits/user_traits.h
nmm0: Maybe we can drop the extra ones for now unless we need this functionality -- if we do need derived or template traits we can add them unconditionally

src/checkpoint/traits/user_traits.h
Matthew-Whitlock: We decided on CamelCase for all structs except ones that are always std::true_type or std::false_type (so we can have the, e.g., `has_trait` and `has_trait_v` names match)

src/checkpoint/traits/user_traits.h
nmm0: I guess that assumption is only used in `_without_trait`. Maybe we could use `_without_trait` outside the class? That would also possibly remove the need for having a separate without_helper thing

src/checkpoint/traits/user_traits.h
```diff
@@ -0,0 +1,176 @@
...
+                          >;
+    template<typename... T>
+    using With = TraitHolder<Trait, Traits..., T...>;
+    //To respect ordering. Could be handy in the future -- disambiguating multiple hooks?
+    template<typename... T>
+    using WithPre = TraitHolder<T..., Trait, Traits...>;
+
+    //Flush out any nested traits (TraitHolder<..., TraitHolder<...>, ...>)
+    //Remove any NoTraits, unless that's the only trait.
+    using BaseTraits = typename TraitHolder<Traits...>::BaseTraits::template WithPre<Trait>;
```
nmm0: Not specializing TraitHolder for empty  might simplify the logic here -- also if we can remove nested traits I guess we might not need this?

nmm0 (approved):
Looks good to me :D

nmm0:
@cz4rs can you re-review?

src/checkpoint/serializers/serializer_ref.h
```diff
@@ -0,0 +1,160 @@
...
+
+#include <type_traits>
+#include <cstdlib>
+#include <memory>
+
+namespace checkpoint {
+
+template<typename SerT, typename UserTraits = UserTraitHolder<>>
+struct SerializerRef
+{
```
cz4rs:
```suggestion
struct SerializerRef {
```

cz4rs (approved):
Looks good to me. The only missing bit would be to integrate it with [the documentation](https://darma-tasking.github.io/checkpoint_docs/html/ckpt_learn_examples.html).

So far all existing examples were contained in .cc files (without headers), but it should be just fine to use two `\snippet`s.

examples/checkpoint_example_user_traits.hpp
```diff
@@ -0,0 +1,109 @@
+#include "checkpoint/checkpoint.h"
```
lifflander: Missing header here

examples/checkpoint_example_user_traits.cc
```diff
@@ -0,0 +1,61 @@
...
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact darma@sandia.gov
+//
+// *****************************************************************************
+//@HEADER
+*/
+#include "checkpoint/checkpoint.h"
```
lifflander, line 46: Add as a documentation snippet

lifflander (changes requested):

examples/checkpoint_example_user_traits.hpp
cz4rs: Use `scripts/check_guards.sh` to add license and header guards here.

cz4rs: Wrong file extension too: use `.h`

examples/checkpoint_example_user_traits.cc
cz4rs: image

cz4rs: image

cz4rs: Docs build seem to be passing now: https://github.com/DARMA-tasking/magistrate/actions/runs/13376630110/job/37357170675?pr=345. I will double-check after this is merged :shrug:

lifflander (approved):
I think this is finally ready to merge! I've completed a final review of it and it looks good.

PR #355, opened 2024-06-24
#354: Bump minimum CMake version to 3.23 and update `find_package_local` to work with `packageName_ROOT`

### LINKED ISSUE #354: Bump `cmake_minimum_required` to 3.23 to match `vt`
TODO: 
- update minimum CMake version to match vt
- We can then update `find_package_local` to support `<PackageName>_ROOT` (we use it for Kokkos and KokkosKernels)

JacobDomagala:
Fixes #354


thearusable (approved):

cmake/load_package.cmake
```diff
@@ -21,7 +21,7 @@ macro(require_pkg_directory pkg_name pkg_user_name)
   endif()
 endmacro(require_pkg_directory)
 
-macro(find_package_local pkg_name pkg_directory pkg_other_name)
+macro(find_package_local pkg_name pkg_other_name)
```
cz4rs: Do we really need the `pkg_other_name` parameter? I think we should be good with just searching for `Kokkos` and `KokkosKernels` is simply repeated anyways.

cmake/load_package.cmake
```diff
@@ -32,63 +32,37 @@ macro(find_package_local pkg_name pkg_directory pkg_other_name)
...
 
-    # Rest of the arguments are potential relative search paths wrt the
-    # ${pkg_directory}
-    set(prefix_args ${ARGN})
-
-    # Default search paths: root, /cmake and /CMake subdirectories
-    list(APPEND prefix_args "/" "/cmake" "/CMake")
-
     # Whether we loaded the package in the following loop with find_package()
     set(${pkg_name}_PACKAGE_LOADED 0)
```
cz4rs: This variable seems obsolete now.

cmake/load_package.cmake
JacobDomagala: For this to work we'd have to update all occurrences of `-Dkokkos_DIR` to `-DKokkos_DIR`. This would be breaking change, but I agree that it should be done.

I don't know the reason we're using lowercase version for Kokkos.

cmake/load_package.cmake
JacobDomagala: Fixed

cmake/load_package.cmake
cz4rs: Alright, lets discuss this at today's meeting.

JacobDomagala: We decided to use `-DKokkos_DIR`

cz4rs (approved):
Looks good to me.

lifflander (approved):

PR #362, opened 2024-07-23
#268: Fix deserialization of polymorphic types when base class is specified as a template parameter.

### LINKED ISSUE #268: This code fails when T is polymorphic
```C++
std::unique_ptr<T> task(new U(1234));
auto ret = checkpoint::serialize<T>(*task);
auto out = checkpoint::deserialize<T>(std::move(ret));
```

@PhilMiller What do you think? Should we support this case?

@nlslatt

thearusable:
Fixes #268

### DIFF
--- a/tests/unit/test_polymorphic.cc
+++ b/tests/unit/test_polymorphic.cc
@@ -3,7 +3,7 @@
 // *****************************************************************************
 //
 //                             test_polymorphic.cc
-//                 DARMA/checkpoint => Serialization Library
+//                 DARMA/magistrate => Serialization Library
 //
 // Copyright 2019 National Technology & Engineering Solutions of Sandia, LLC
 // (NTESS). Under the terms of Contract DE-NA0003525 with NTESS, the U.S.
--- a/tests/unit/test_polymorphic.cc
+++ b/tests/unit/test_polymorphic.cc
@@ -112,6 +112,8 @@ void testPolymorphicTypes(int val) {
   auto out = checkpoint::deserialize<Base>(std::move(ret));
 
   EXPECT_TRUE(nullptr != out);
+  EXPECT_EQ(typeid(*task), typeid(*out));
+  EXPECT_TRUE(nullptr != dynamic_cast<Derived*>(out.get()));
   EXPECT_EQ(val, out->getVal());
 }
 
--- a/tests/unit/test_polymorphic.cc
+++ b/tests/unit/test_polymorphic.cc
@@ -115,7 +115,7 @@ void testPolymorphicTypes(int val) {
   EXPECT_EQ(val, out->getVal());
 }
 
-TEST_F(TestPolymorphic, test_polumorphic_type) {
+TEST_F(TestPolymorphic, test_polymorphic_type) {
   testPolymorphicTypes<Derived2, Derived2>(5);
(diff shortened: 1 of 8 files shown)


tests/unit/test_virtual_serialize.cc
```diff
@@ -611,6 +611,28 @@ INSTANTIATE_TYPED_TEST_CASE_P(
...
+
+using TestDeserializationFromBase = TestHarness;
+
+template<typename Base, typename Derived>
+void testDeserializationFromBase() {
+  std::unique_ptr<Base> task(new Derived(TEST_CONSTRUCT{}));
+  auto ret = checkpoint::serialize<Base>(*task);
+  auto out = checkpoint::deserialize<Base>(std::move(ret));
+
+  EXPECT_TRUE(nullptr != out);
```
PhilMiller, line 626: There needs to be some explicit test here, rather than within `check()`, that the deserialization really did produce an instance of `Derived` - either that `dynamic_cast` returns non-`nullptr`, or that `getID()` returns the right enumerator value. Otherwise, the call `out->check()` could dispatch to the `Base` implementation, and mistakenly pass the test.

tests/unit/test_virtual_serialize.cc
```diff
@@ -611,6 +611,28 @@ INSTANTIATE_TYPED_TEST_CASE_P(
...
+  out->check();
+}
+
+TEST_F(TestDeserializationFromBase, test_deserilization_from_base) {
```
PhilMiller:
```suggestion
TEST_F(TestDeserializationFromBase, test_deserialization_from_base) {
```
Typo fix there

tests/unit/test_virtual_serialize.cc
PhilMiller: It looks like the new `test_polymorphic.cc` probably covers this concern. I still think this test should be made comprehensive, though, so as not to accept integration of an incomplete test.

PhilMiller:
I'll wait to look deeper at the implementation until the tests are passing.

tests/unit/test_virtual_serialize.cc
thearusable: Fixed.

tests/unit/test_virtual_serialize.cc
thearusable: Right! I added a check for the id returned by `getID()`

tests/unit/test_polymorphic.cc
```diff
@@ -0,0 +1,127 @@
...
+  }
+};
+
+template<typename Base, typename Derived>
+void testPolymorphicTypes(int val) {
+  std::unique_ptr<Base> task(new Derived(val));
+  auto ret = checkpoint::serialize(*task);
+  auto out = checkpoint::deserialize<Base>(std::move(ret));
+
+  EXPECT_TRUE(nullptr != out);
```
lifflander, line 114: I think we should probably include an ID check here to ensure we are getting the right type instead of just checking that it's not null and the output value is correct.

lifflander (changes requested):

tests/unit/test_polymorphic.cc
thearusable: I added the check for typeid.

Issue #337, opened 2024-05-07
Fix checkpoint CI documentation push failing authentication


lifflander:

lifflander:
This has been fixed by updating the token.

PR #393, opened 2025-07-16
#392: Clean up `docker-compose.yml`

### LINKED ISSUE #392: Clean up docker-compose.yml
Reduce redundancy, remove unused sections. Leverage the workflow images (variants instead of separate alpine, openmpi, docs entries).

JacobDomagala:
Fixes #392


.github/workflows/build-docker-images.yml
```diff
@@ -43,6 +44,7 @@ jobs:
           key: ${{ matrix.target }}-${{ github.ref }}-${{ github.sha }}
           restore-keys: |
             ${{ matrix.target }}-${{ github.ref }}
+            ${{ matrix.target }}-
```
lifflander, line 47: Is the ref not needed for the restore-keys?

lifflander (approved):
Looks good

.github/workflows/build-docker-images.yml
JacobDomagala: `github-cache` uses restore keys in the order they are declared. This change allows the cache to use the `${{ matrix.target }}-` restore key if all previous keys result in cache misses. 

We can reuse the cache from `develop` branch generated by the same bake target, when there is no cache matching commit or branch.

docker-compose.yml
```diff
@@ -101,59 +63,34 @@ x-checkpointopts: &checkpointopts
...
+      <<: [*ccache, *magistrateopts]
+    volumes: &ubuntu-volumes
+      - .:/magistrate:delegated
+      - ${CACHE}${ARCH}-ubuntu-${UBUNTU}-${COMPILER}-cache:/build:delegated
 
   ##############################################################################
-  # Build documentation for Checkpoint in the container on ubuntu platform from
+  # Build documentation for magistrate in the container on ubuntu platform from
   # container baseline.
   ubuntu-docs:
```
cz4rs: We want to migrate docs build to bake too, now might be the good time to do it.

Note that the documentation is currently failing on `develop` with:
```
+ git config --global user.email jliffla@sandia.gov
+ git config --global user.name 'Jonathan Lifflander'
+ git add checkpoint_docs
+ git commit --allow-empty -m 'Update magistrate_docs (auto-build)'
[master 9f3c37c] Update magistrate_docs (auto-build)
 1 file changed, 2 insertions(+), 2 deletions(-)
 rewrite checkpoint_docs/html/searchdata-v2.js (97%)
+ git push origin master
fatal: could not read Password for 'https://***@github.com': No such device or address
```

cz4rs (approved):
Looks good to me.

JacobDomagala:
Will try to use docs build with bake - moving to draft.

JacobDomagala:
This should be good to go once we update `GH_PAT` token. Looks like it expired (last documented update was roughly 1 year ago - https://github.com/DARMA-tasking/magistrate/issues/337)

~I will change the trigger to `push: develop` after testing~

.github/workflows/build-docs.yml
```diff
@@ -1,27 +1,19 @@
-name: Build Documentation
+name: magistrate-docs
```
cz4rs: I think we could have it running from `build-docker-images.yml` and remove this file completely. See my attempt in vt: https://github.com/DARMA-tasking/vt/pull/2456/commits/3181e177568a7de43eeba91becb13a51a590c47e.
We'd just have to add restriction for running only on `develop` - we could use your previous approach from spack PR.

docker-bake.hcl
```diff
@@ -72,6 +74,14 @@ function "magistrate_code_coverage" {
...
 }
 
+function "magistrate_suffixes" {
+  params = [item]
+  result = join("", compact([
+    magistrate_build_against_vt(item) == 1 ? "-vt" : "",
+    magistrate_docs(item) == 1 ? "-docs" : ""
+  ]))
+}
+
```
cz4rs: Can you check the logic for `variant` in workflows `docker-bake.hcl`? I'd prefer something closer to that for consistency and having the suffix right in the build definition.

cz4rs: (I'll be adding something similar in `vt` as well)

docker-bake.hcl
```diff
@@ -139,6 +152,10 @@ target "magistrate-build-all" {
         magistrate_code_coverage = 1,
         magistrate_serialization_error_checking = 0
       },
+      {
+        image = "amd64-ubuntu-20.04-gcc-9-cpp",
+        magistrate_docs = 1
```
cz4rs: Something like `variant = "docs"` here.

ci/build_cpp.sh
```diff
@@ -57,21 +51,26 @@ if test "${MAGISTRATE_DOXYGEN_ENABLED:-0}" -eq 1
 then
     MCSS=$PWD/m.css
     GHPAGE=$PWD/DARMA-tasking.github.io
-    git clone --depth=1 "https://${token}@github.com/DARMA-tasking/DARMA-tasking.github.io"
+
+    git clone --depth=1 "https://x-access-token:${TOKEN}@github.com/DARMA-tasking/DARMA-tasking.github.io"
```
cz4rs: Shouldn't this be `GITHUB_TOKEN`?
```suggestion
    git clone --depth=1 "https://x-access-token:${GITHUB_TOKEN}@github.com/DARMA-tasking/DARMA-tasking.github.io"
```

ci/build_cpp.sh
```diff
@@ -57,21 +51,26 @@ if test "${MAGISTRATE_DOXYGEN_ENABLED:-0}" -eq 1
...
-    mv docs "$CKPT_NAME"
-    cp  -R "$CKPT_NAME" "$GHPAGE"
-    cd "$GHPAGE"
-    git config --global user.email "jliffla@sandia.gov"
-    git config --global user.name "Jonathan Lifflander"
-    git add "$CKPT_NAME"
-    git commit --allow-empty -m "Update magistrate_docs (auto-build)"
-    git push origin master
+
+    if test "${GIT_BRANCH:-}" = "develop"
```
cz4rs, line 62: Does it work with `git branch --show-current`? We could skip passing `GIT_BRANCH` this way.

cz4rs (changes requested):
See comments and questions posted.

docker-bake.hcl
cz4rs: See `variant` and `target_suffix` in https://github.com/DARMA-tasking/vt/pull/2456/commits/cb0dd47b18cda0861516497c6af432c86aef8491.

ci/build_cpp.sh
JacobDomagala: We have `.git` in `dockerignore`, there's no information about commit/branch etc.

docker-bake.hcl
JacobDomagala: Sounds good

ci/build_cpp.sh
JacobDomagala: Good catch! I renamed that token quite few times, and since this part of the code only runs on develop, I wasn't able to notice that :D

ci/build_cpp.sh
JacobDomagala: This is also the reason why our `codecoverage` report was lacking, we were using it inside isolated docker image.

cz4rs (approved):
LGTM

lifflander (approved):
Looks good

PR #395, opened 2025-07-23
#394: Remove old workflows and fix code coverage build

### LINKED ISSUE #394: [CI] Remove old docker builds
Once we have docker bake working (https://github.com/DARMA-tasking/magistrate/pull/390) we can remove the old workflows

JacobDomagala:
Fixes #394


README.md
```diff
@@ -1,19 +1,11 @@
...
-[![clang-14 ubuntu mpich](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-clang-14-ubuntu-mpich.yml/badge.svg)](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-clang-14-ubuntu-mpich.yml)
-[![gcc-9 ubuntu mpich](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-gcc-9-ubuntu-mpich.yml/badge.svg)](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-gcc-9-ubuntu-mpich.yml)
-[![gcc-10 ubuntu mpich](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-gcc-10-ubuntu-mpich.yml/badge.svg)](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-gcc-10-ubuntu-mpich.yml)
-[![nvcc-11 ubuntu mpich](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-nvcc-11-ubuntu-mpich.yml/badge.svg)](https://github.com/DARMA-tasking/checkpoint/actions/workflows/dockerimage-nvcc-11-ubuntu-mpich.yml)
+[![CI](https://github.com/DARMA-tasking/magistrate/actions/workflows/build-docker-images.yml/badge.svg)](https://github.com/DARMA-tasking/magistrate/actions/workflows/build-docker-images.yml)
```
JacobDomagala, line 6: GitHub Actions does not natively support per-matrix-job badges. The default badge reflects the overall workflow status, not individual matrix variations

lifflander: :( Ok, I guess that's the best we can do then.

lifflander:
Before we go through with the removal, are you confident that the new workflows are identical in terms of cmake configuration to the old ones?

JacobDomagala:
Yes, I went through them manually, and that's how I noticed gcc-9 build wasn't actually using code coverage (despite it having `magistrate_code_coverage = 1` in the bake config file)

README.md
cz4rs: :/ I wonder if there's any reliable workaround for this. @JacobDomagala Can you create a low priority issue for this after merging?

cz4rs (approved):
Looks good to me.

README.md
JacobDomagala: Since this is something that vt will also want to use, I think we should create a custom github action that would generate the badge based on the name and build output (success/failure). You would use it as follows:

```
- name: Generate build badge
  if: always()
  uses: Darma-tasking/badge-generator@master
  with:
    - name: ${{ matrix.target }}
    - result: ${{ steps.docker-build.outcome }}
```


Badge generation is quite easy; you can use shields.io (example: https://img.shields.io/badge/custom_name-green.svg - click on it). We're currently using this website to generate VT's build-stats workflow badge.
The only downside of this approach is the same reason it's so effective: you have to modify the URL itself to change the badge. This makes it quite challenging to use statically inside a README.

The way we handle it for build-stats is that we store it in VT's wiki pages and then reference it in VT's README as follows:

`[](https://github.com/DARMA-tasking/vt/wiki/Build-Stats)`


## Static badge
First approach would be the same as we do for `build-stats`. The action would store the generated badges inside its wiki pages, and all we would have to do is to reference it inside README like so:

```
[![](https://github.com/DARMA-tasking/badge-generator/wiki/magistrate-build-amd64-ubuntu-22-04-clang-14-cpp)](https://github.com/DARMA-tasking/magistrate/actions/workflows/build-docker-images.yml?query=branch%3Adevelop)
```

Result (clickable): 
[](https://github.com/DARMA-tasking/magistrate/actions/workflows/build-docker-images.yml?query=branch%3Adevelop)

Small issue with this approach is that when you click on the badge, it will redirect you to the history of the builds, you then have to manually select the failing one. 

----

## Dynamic badge

Other way of handling this would be to have an action that runs after all matrix builds have finished and generates new badges, and finally updates `README`. We no longer need to store the badges to wiki pages as we will use `shields.io` directly. 

This provides a better user experience - you click on the badge and it redirects you to the actual build. The obvious downside is that we will have an extra commit after merging to `develop`.

Result (clickable): 
[](https://github.com/DARMA-tasking/magistrate/actions/runs/16423361169/job/46407456383)

.github/workflows/build-docker-images.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: Build docker images
```
cz4rs: Nitpicking: I actually like `CI` here, because that's pretty much the gist of it. We run almost all of our CI through bake, so the fact that it builds docker images might be considered an implementation detail :wink:

README.md
```diff
@@ -1,19 +1,12 @@
-# *magistrate* => serialization and checkpointing library
+<div align="center">
```
cz4rs, line 1: image

Looks good to me, we can get back to adding detailed badges in a separate PR.

JacobDomagala:
Moving this to draft as codecov issue is not yet fully resolved (new token works, but we have .git directory inside dockerignore and codecov is missing git related information)

cz4rs (approved):

README.md
JacobDomagala: Created https://github.com/DARMA-tasking/magistrate/issues/396

.github/workflows/build-docker-images.yml
JacobDomagala: :heavy_check_mark:

.github/workflows/build-docker-images.yml
```diff
@@ -72,3 +72,25 @@ jobs:
...
+          FROM ubuntu:latest
+          RUN --mount=type=cache,id=BUILD-${{ matrix.target }},target=/build/checkpoint \
+          mkdir -p /checkpoint \
+          && cp -p -R /build/checkpoint/. /checkpoint/ || true
+          " >> tmp_dockerfile
+
+          docker buildx build --builder ${{ steps.setup-buildx.outputs.name }} -f tmp_dockerfile --output type=local,dest=build .
+
+      - name: Upload coverage
+        if: ${{ matrix.target == 'magistrate-build-amd64-ubuntu-20-04-gcc-9-cpp' }}
```
cz4rs, line 92: In theory we could do `docker buildx bake --print ${{ matrix.target }} | grep MAGISTRATE_CODE_COVERAGE` etc., but I think having `gcc-9` hardcoded here is fine.

cz4rs (approved):
Looks great :)

.github/workflows/dockerimage-clang-10-ubuntu-mpich.yml
```diff
@@ -1,76 +0,0 @@
-###############################################################################
```
cz4rs, line 1: I think you can remove `generate-workflow.pl`, `workflow-template.yml` and `workflows.ini`.

lifflander (approved):
Looks good to me

PR #402, opened 2025-09-24
#401: fix memory leaks, improve design around in-place allocs

### LINKED ISSUE #401: Destructors not firing properly on in-place new allocations causing memory leaks
This is most clearly found by de-serializing a map-like type `std::map<std::string, int>` with a long (non-SSO) string that allocates. This happens to trigger this case as the `std::move` is actually a copy since the underlying type serialized is `std::pair<std::string const, int>`, which is not movable (which is a secondary performance concern).

lifflander:
Fixes #401

### DIFF
--- a/src/checkpoint/container/container_serialize.h
+++ b/src/checkpoint/container/container_serialize.h
@@ -90,24 +90,39 @@ serializeContainerCapacity(Serializer& s, ContainerT& cont) {
 
 template <typename Serializer, typename ContainerT>
 inline void serializeContainerElems(Serializer& s, ContainerT& cont) {
-  // JL: Without error checking we can simply iterate and seriialize
-  // these. However, since we are changing the type to remove const from
-  // std::pair<X const, Y>, error checking detects the mis-alignment. Here we
-  // detect and massage the types to match what is happening in the
-  // serialization path based on the how the value_type comes out. For example,
-  // for std::set, the value_type is always const because elements can't be
-  // modified, so we have to const_cast this out.
+  // Without error checking, we can easily iterate and serialize by iterating
+  // through. However, since we are modifying the type to eliminate const from
+  // std::pair<X const, Y>, error checking identifies the type mismatch. Here,
+  // we adjust the types to align with the serialization process based on the
+  // resulting value_type. For instance, in the case of std::set, the value_type
+  // is always const because the elements cannot be altered, necessitating the
+  // use of const_cast to remove the const qualifier.
 
+#if defined(SERIALIZATION_ERROR_CHECKING)
   using ValueT = typename detail::get_value_type<ContainerT>::value_type;
   for (auto& elm : cont) {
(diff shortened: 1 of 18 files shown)


lifflander:
The error checking detects the type difference between the `std::pair<const T, U>` and the new transformed type `std::pair<T, U>` and now fails...

src/checkpoint/container/container_serialize.h
```diff
@@ -73,8 +90,23 @@ serializeContainerCapacity(Serializer& s, ContainerT& cont) {
...
+  // modified, so we have to const_cast this out.
+
+  using ValueT = typename detail::get_value_type<ContainerT>::value_type;
+  for (auto& elm : cont) {
+    if constexpr (std::is_same<ValueT&, decltype(elm)>::value) {
+      s | elm;
+    } else if constexpr (std::is_same<ValueT const&, decltype(elm)>::value) {
+      s | const_cast<ValueT&>(elm);
+    } else {
+      s | reinterpret_cast<ValueT&>(elm);
```
PhilMiller, line 118: Should this case be here? Should it be a `static_assert(dependent_false<ValueT>)`? It's not clear that a `reinterpret_cast` is really the right thing. Maybe comment it if it actually is.

src/checkpoint/dispatch/reconstructor.h
```diff
@@ -49,10 +49,85 @@
...
+/**
+ * \struct InPlaceWrapper
+ *
+ * \brief Wrapper class for a in-place allocation which may not have a paired
+ * destructor. If a user has a pre-existing allocation which magistrate
+ * de-serializes into, it may never be destructed as it is copied or moved
+ * around. This type wraps that allocation and forces it to be destroyed or
+ * explicity requires \c transferOwnership to be called to document that the
+ * user is now responsible for destructing it (normally when the type if put in
+ * a managed pointer, e.g., a std::unique_ptr with a standard deleter.
```
PhilMiller: This comment seems to trail off, or something. Parenthetical is unclosed, and the grammar doesn't make sense.

lifflander: Good call, I didn't finish writing this one!

src/checkpoint/container/container_serialize.h
lifflander: I've added comments. Do they make sense? Do you see a better way of handling this case with `std::pair<T const, U>` -> `std::pair<T, U>`

PhilMiller: Yeah, that makes sense now. I was mostly confused about the last case, but it's good to have them all described like this.

Is there a reason to have the `#if / #else`? It doesn't seem like the error checking case adds any overhead

lifflander: No, I actually added that just for cleanliness to made it clear that it's not a correctness requirement to adjust the types. Just purely for error checking (aligning the types).

nlslatt (approved):

PR #408, opened 2026-01-07
add option to disable export of magistrate targets

rppawlo:
This commit allows trilinos to build magistrate as an official package. This will allow trilinos code to use the serialization tools for checkpointing.

If we build magistrate nested in trilinos as an official trilinos package, we get an error for exporting the same magistrate targets multiple times. The tribits build system is handling the target export for all packages. This PR allows us to disable the magistrate exports to build and install with trilinos.


rppawlo:
@lifflander - would you mind reviewing this change? This is all that was needed to get magistrate building as part of Trilinos. Its for the use case where kokkos and kokkos-kernels are built as part of trilinos. In this case, we need to insert the build of magistrate in between trilinos packages. There is a simple mechanism to do this in trilinos, but it requires control of the config file generation.

lifflander (approved):

lifflander:
Looks good to me. I just signed the commit so it's verified by rebasing.

